PR #1, opened 2018-05-23
Prepare repository for being published

pronebird:
This PR adds licenses and documentation in `lib.rs`, `README.md` generated with `cargo readme` and refines and adds new examples.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)


LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
...
+   To apply the Apache License to your work, attach the following
+   boilerplate notice, with the fields enclosed by brackets "[]"
+   replaced with your own identifying information. (Don't include
+   the brackets!)  The text should be enclosed in the appropriate
+   comment syntax for the file format. We also recommend that a
+   file or class name and description of purpose be included on the
+   same "printed page" as the copyright notice for easier
+   identification within third-party archives.
+
+Copyright [yyyy] [name of copyright owner]
```
pronebird: @faern do we have to fix this? I copied this from `jsonrpc-client-rs`

examples/install_service.rs
```diff
@@ -0,0 +1,35 @@
...
+    let service_manager = ServiceManager::local_computer(None::<&str>, manager_access).unwrap();
+    let service_info = ServiceInfo {
+        name: OsString::from("ping_service"),
+        display_name: OsString::from("Ping service"),
+        service_type: ServiceType::OwnProcess,
+        start_type: ServiceStartType::OnDemand,
+        error_control: ServiceErrorControl::Normal,
+        executable_path: ::std::env::current_exe()
+            .unwrap()
+            .with_file_name("ping_service.exe"),
```
pronebird: This example registers the other example as a service, so all three examples we have a connected.

examples/ping_service.rs
```diff
@@ -0,0 +1,137 @@
+// Ping service example.
```
pronebird: Way better example to get started with.

LICENSE-APACHE
faern: Hmm. It's probably good to insert values there indeed. Should probably be done in the other libraries as well, but maybe not very high priority. `2018` and `Amagicom AB` would be correct I guess.

examples/install_service.rs
faern: Do they have to be? It's nice with examples that are small and isolated and can be studied and understood without having to grasp something larger.

faern: I read the examples now. I feel they are indeed isolated and separated. One can understand `install_service.rs` without understanding `ping_service`. So it's not really a problem IMO.

But it might be good with a comment or two somewhere along the lines of "Installs a service, here we use the other example binary, `ping_service`, to showcase the installation, but it can be replaced with any binary that the service will be based on" or similar. Because as a new reader it can be hard to know what code is just "here is an example of what value one could put here" and "This value must be exactly this for it to work", and it's always good to explain those things IMO.

faern:
Review status: 6 of 10 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/lib.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDDart-AwKTLgkqdOu:-LDDDart-AwKTLgkqdOv:b-5i8ctx) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L14)):*
> ```Rust
> //!
> //! This section describes the steps to create a windows service, for complete source code take a
> //! look at examples folder.
> ```

I think there is risk to confuse "create a windows service" with "install a windows service". Might be beneficial for the understanding to be clear that when we say "implementing a windows service" we mean writing a program *in Rust* that *runs as a system service*.

For the same reason might be beneficial to point out that this corresponds to the `ping_service` example, so it's not confused with installing or uninstalling services.

---

*[src/lib.rs, line 18 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDE8eYDFUjG0k53Teu:-LDDE8eYDFUjG0k53Tev:b-5kvxxo) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L18)):*
> ```Rust
> //! ## Basics
> //!
> //! Each windows service has to implement a `service_main` function and register it with the system
> ```

`service_main` is just an example, right? I don't think it should be mentioned here, because it sounds like the function must have that name. Instead of a function name I suggest you write out the signature of the function that the binary must export in order to be able to run as a service. And then say that the `define_windows_service!` macro generate such a function for you with any name the user want. Would be beneficial to include more links as well when you mention macros and similar stuff that is part of the public API of this crate.

https://rust-lang-nursery.github.io/api-guidelines/documentation.html#prose-contains-hyperlinks-to-relevant-things-c-link

---

*[src/lib.rs, line 28 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDEzH59g9Yj8o6UP8z:-LDDEzH59g9Yj8o6UP9-:b-ah79sj) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L28)):*
> ```Rust
> //! use windows_service::service_dispatcher;
> //!
> //! define_windows_service!(service_main, handle_service_main);
> ```

I don't think `service_main` and `handle_service_main` are ideal names for these functions. Just when reading the two names I have no idea how they differ from each other. Something like `exported_ffi_service_main` and `my_service_main` might be a little more clear, but that's the best substitution I came up with on the spot.

---

*[src/lib.rs, line 31 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDFQGE6D0RxGccYS1s:-LDDFQGE6D0RxGccYS1t:bchkw6u) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L31)):*
> ```Rust
> //!
> //! fn handle_service_main(arguments: Vec<OsString>) {
> //!     // Do service work
> ```

Maybe be a little more clear that this is "the main entry point of the service, where execution will start after the call to `start_dispatcher`".. Or something similar. 

---

*[src/lib.rs, line 36 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDFfiA4jUQbvzD0srs:-LDDFfiA4jUQbvzD0srt:b-5dpl18) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L36)):*
> ```Rust
> //! fn main() {
> //!     // Register generated `service_main` with the system and start the service blocking main
> //!     // thread until the service is stopped.
> ```

Since you are already talking about threads here it might be nice to also add that the `service_main` function will be called from another thread?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: 6 of 10 files reviewed at latest revision, 15 unresolved discussions.

---

*[src/lib.rs, line 43 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDGcpJ0Ac0C2MxLnCE:-LDDGcpK7w0ERslT98CR:b-1w2bx7) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L43)):*
> ```Rust
> //! ## Handling service events
> //!
> //! The first thing that windows service should do early in its lifecycle is to subscribe for
> ```

that **a** windows service

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: 6 of 10 files reviewed at latest revision, 16 unresolved discussions.

---

*[src/lib.rs, line 76 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDHXzXDTliyBoQVR4u:-LDDHXzYCurqynuKz3BH:b-x93txj) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L76)):*
> ```Rust
> //!
> //! The service status handle (`service_control_handler::ServiceStatusHandle`) is issued upon
> //! successful event handler registration (see `service_control_handler::register_control_handler`)
> ```

A lot of the function names and paths in this crate has very long names and are a bit redundant. `service_control_handler::register_control_handler` has `control_handler` in it twice. Just `service_control_handler::register` would be a lot easier to read for example, that one is registering a control_handler feels given from the context.

Even better if `register` was a constructor on `ServiceStatusHandle` since that is the type the function returns?

Sorry for coming with this type of feedback so late. But I have realized that even after a lot of reviewing in this codebase I have very little grasp about what exists and what does what. Feels like that is mostly because there are so many very long names.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

examples/install_service.rs
pronebird: OK. that makes sense. I will comment on that. Apparently I didn't want to bundle installation into the service example since one can do that on their own if needed + it saves a few dozen LOC and makes examples easier to follow.

LICENSE-APACHE
pronebird: Done

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 60 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD6uzD4gdlz3cv8yka:-LDGtZwB-2no1FZpE9FU:b6x532s) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L60)):*

Good and informative comment :+1:
</blockquote></details>

Thanks :+1: 

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 69 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD75ti8pnAQKiArtFw:-LDGtcSLDrYemYtwueLK:b-hgrtop) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L69)):*

s/NO_ERROR/NoError? Maybe it's NO_ERROR underneath, but that is not the level we are exposing to the user here.
</blockquote></details>

Old comment from the times when I used the raw constant directly. Probably still exists somewhere in mullvad-daemon's source :laughing: 
Fixed it here.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 93 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD7eBnCSsyC2oy215X:-LDGts6i8UKUxtsc6Zbv:b53o8wi) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L93)):*

As a reader it feels strange that I'm sending the `Running` status and at the same time send an `exit_code`. If it's true that the `exit_code` has to be present on all `set_service_status` but that it's ignored, or should be zero, on all non-exited statuses then I think it would be nice to say so in a comment about this field.
</blockquote></details>

As per docs:

1. `checkpoint` has to be set to `0` for non-pending states.
2. `wait_hint` is ignored for non-pending states.
3. `exit_code` is ignored for the most of the time except maybe `Stopped`.

We should probably point people to MSDN at this point or provide better means to manage the service status.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 95 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD80h51w_ik3-6c3NA:-LDGzaCG-ugNp9KmM3IP:b-o3xnm4) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L95)):*

Both the `checkpoint` and the `wait_hint` are not obvious what they are from just their names. Maybe a short description about them, or just a reference to other documentation about these fields?
</blockquote></details>

Yeah, MSDN link is missing for sure.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 106 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8qCdCtzARIBukUdc:-LDGzzItCbb19-ZefTeG:bvycna2) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L106)):*

Don't you want to bind the socket outside the loop? So you don't allocate and bind a new socket every second?
</blockquote></details>

I wasn't sure if that would work in case there is no listener on the port.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 122 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8FGU7_xM-eGRPCZg:-LDH-_3h3GGKKxtLNkTQ:beiv018) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L122)):*

Why do you need to spawn a separate thread if the only thing the current one is waiting for that other thread anyway. Could you not just run the code in that thread here instead of waiting?
</blockquote></details>

Certainly we can run the loop in `service_main`

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/uninstall_service.rs, line 23 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDDFVv0AWeHUSiJUh5:-LDH-pWE0HUIbguj8LV4:b-9i33cz) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/uninstall_service.rs#L23)):*

Would it be possible that instead of just waiting for a hardcoded amount of time, instead query the status in a loop and wait until it's `ServiceStatus::Stopped`
</blockquote></details>

That's done for simplicity, the one can do the loop of course.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDDart-AwKTLgkqdOu:-LDH0Jpw0yxl6QXWi-ni:bto4mba) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L14)):*

I think there is risk to confuse "create a windows service" with "install a windows service". Might be beneficial for the understanding to be clear that when we say "implementing a windows service" we mean writing a program *in Rust* that *runs as a system service*.

For the same reason might be beneficial to point out that this corresponds to the `ping_service` example, so it's not confused with installing or uninstalling services.
</blockquote></details>

This is more of a step-by-step guide, it's not related to `ping_service` example. I'll reword that part to be clearer.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 18 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDE8eYDFUjG0k53Teu:-LDH70oy1IH_EXojKVzJ:bdq6ka0) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L18)):*

`service_main` is just an example, right? I don't think it should be mentioned here, because it sounds like the function must have that name. Instead of a function name I suggest you write out the signature of the function that the binary must export in order to be able to run as a service. And then say that the `define_windows_service!` macro generate such a function for you with any name the user want. Would be beneficial to include more links as well when you mention macros and similar stuff that is part of the public API of this crate.

https://rust-lang-nursery.github.io/api-guidelines/documentation.html#prose-contains-hyperlinks-to-relevant-things-c-link
</blockquote></details>

Oh good point, I didn't know hyperlinks work!

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 18 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDE8eYDFUjG0k53Teu:-LDH7KRO93nBykodFNxR:bb6un1j) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L18)):*

`service_main` is just an example, right? I don't think it should be mentioned here, because it sounds like the function must have that name. Instead of a function name I suggest you write out the signature of the function that the binary must export in order to be able to run as a service. And then say that the `define_windows_service!` macro generate such a function for you with any name the user want. Would be beneficial to include more links as well when you mention macros and similar stuff that is part of the public API of this crate.

https://rust-lang-nursery.github.io/api-guidelines/documentation.html#prose-contains-hyperlinks-to-relevant-things-c-link
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 104 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8SJC84OtEdjdhL5k:-LDHH4RuBM1hOtP7ME4q:brc3nkk) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L104)):*

I think it would increase the readability and how fast one can grasp what this file does if the IP and port were defined in consts high up in the file. So one immediately after reading the comment at the start of the file can see the constants that actually control the "ping" target. The string to be sent in the packet could very well be defined up there as well IMO.
</blockquote></details>

I don't think it would improve anything. I think everything in one place is easier to follow than well structured code when it comes to examples.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 106 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8qCdCtzARIBukUdc:-LDHI7Fb0VB0kU4rN7Gw:bvvhlr2) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L106)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

I wasn't sure if that would work in case there is no listener on the port.
</blockquote></details>

Actually UDP doesn't care. Great suggestion, makes ncat work flawlessly too. Previous ncat would only accept 1 connection.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 106 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8qCdCtzARIBukUdc:-LDHIatWByRzot_mnaNY:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L106)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Actually UDP doesn't care. Great suggestion, makes ncat work flawlessly too. Previous ncat would only accept 1 connection.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 113 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD98Y92TdjIflsVpkA:-LDHIbto49gmUyOkVpZq:b-yspirz) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L113)):*

Would you not want to `break` instead of `return`? Kind of the same thing in this particular scenario, but the comment says you break the loop :)
</blockquote></details>

The one can break the loop by either `break` or `return`. Apparently the difference is in the extent of the used keyword. I changed that to `break` if you think it's better.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 31 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDFQGE6D0RxGccYS1s:-LDHIyW46PLJE5zTRTof:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L31)):*

Maybe be a little more clear that this is "the main entry point of the service, where execution will start after the call to `start_dispatcher`".. Or something similar. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 36 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDFfiA4jUQbvzD0srs:-LDHJ1HZ-MaKTQf5q2sj:b-at26mq) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L36)):*

Since you are already talking about threads here it might be nice to also add that the `service_main` function will be called from another thread?
</blockquote></details>

Mentioned this in `handle_service_main` above

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 43 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDGcpJ0Ac0C2MxLnCE:-LDHJHEW5yS-QrtC3tWX:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L43)):*

that **a** windows service
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 10 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 111 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8wy3D1EaJzkvbvic:-LDHLwwp6yzAfBjQKNSw:bid6ts9) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L111)):*

This `Duration` feels like something that would benefit from also being a constant high up in the file. So all parameters controlling the execution are visible from the get go and easy to change without going through the code.
</blockquote></details>

I seriously don't see any benefit of polishing the example code that much. It's not like anyone is going to change that and release this code...

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 11 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 28 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDEzH59g9Yj8o6UP8z:-LDHNDtL7dz8wrHVujBl:b-6iafkh) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L28)):*

I don't think `service_main` and `handle_service_main` are ideal names for these functions. Just when reading the two names I have no idea how they differ from each other. Something like `exported_ffi_service_main` and `my_service_main` might be a little more clear, but that's the best substitution I came up with on the spot.
</blockquote></details>

How about `ffi_service_main` and `my_service_main`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 4 of 11 files reviewed at latest revision, 17 unresolved discussions.

---

*[examples/ping_service.rs, line 93 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD7eBnCSsyC2oy215X:-LDHNvE0Exg7LE5SbbNC:b821lga) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L93)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

As per docs:

1. `checkpoint` has to be set to `0` for non-pending states.
2. `wait_hint` is ignored for non-pending states.
3. `exit_code` is ignored for the most of the time except maybe `Stopped`.

We should probably point people to MSDN at this point or provide better means to manage the service status.
</blockquote></details>

I put the link to MSDN in the comment above that line.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Reviewed 1 of 1 files at r2, 1 of 7 files at r3.
Review status: 6 of 13 files reviewed at latest revision, 6 unresolved discussions.

---

*[examples/ping_service.rs, line 104 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8SJC84OtEdjdhL5k:-LDMG7-LDcNxQSZOLTFA:b-b8har0) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L104)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

I don't think it would improve anything. I think everything in one place is easier to follow than well structured code when it comes to examples.
</blockquote></details>

But the difference is essentially values vs code. I would argue that the IPs are values, not code. I agree that code that belongs together should be in the same place and well structured. But the IPs should basically be arguments sent in to the code so it knows where it should do what it's supposed to do :)

---

*[examples/ping_service.rs, line 111 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8wy3D1EaJzkvbvic:-LDMHEzNDqakDLWfSsZp:b-4ckcnj) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L111)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

I seriously don't see any benefit of polishing the example code that much. It's not like anyone is going to change that and release this code...
</blockquote></details>

Release: No. Change: Maybe, in order to learn by playing around. But yeah, I agree this is much less important than the IPs for understanding.

---

*[examples/ping_service.rs, line 113 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD98Y92TdjIflsVpkA:-LDMHYyE5eyk1NRWVlzA:b8ojffr) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L113)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

The one can break the loop by either `break` or `return`. Apparently the difference is in the extent of the used keyword. I changed that to `break` if you think it's better.
</blockquote></details>

Hmm, what? They are not the same. `return` would return from the entire function, not jump to right after the loop. Which was basically the same before when the loop was inside a closure, because then `return` would return from the closure and `break` would jump out of the loop which would then directly return from it.

---

*[src/lib.rs, line 28 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDEzH59g9Yj8o6UP8z:-LDMIWg8C_h3-7a3_t7q:b73cfk9) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L28)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

How about `ffi_service_main` and `my_service_main`?
</blockquote></details>

Sure. That is much better IMO.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 6 of 14 files reviewed at latest revision, 3 unresolved discussions.

---

*[examples/ping_service.rs, line 113 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD98Y92TdjIflsVpkA:-LDaoYvXDXRziGFEKlZ_:b7o0622) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L113)):*

Hmm, what? They are not the same. `return` would return from the entire function, not jump to right after the loop. Which was basically the same before when the loop was inside a closure, because then `return` would return from the closure and `break` would jump out of the loop which would then directly return from it.
</blockquote></details>

What I meant is that each keyword `break` and `return` break the loop to certain extent. 

`break` only exits the loop and continues with the instructions after the loop.

`return` exits the loop in that particular case and the function altogether. 

So in a general sense both keywords "break" the loop. :hammer: 

Let's move on.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 6 of 14 files reviewed at latest revision, 3 unresolved discussions.

---

*[examples/ping_service.rs, line 104 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8SJC84OtEdjdhL5k:-LDawQgr1oT8XVbY7KLY:bd17dd8) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L104)):*

But the difference is essentially values vs code. I would argue that the IPs are values, not code. I agree that code that belongs together should be in the same place and well structured. But the IPs should basically be arguments sent in to the code so it knows where it should do what it's supposed to do :)
</blockquote></details>

Alright man, but the best I can do is to move the localhost as array to static storage since `IpAddr` cannot be statically allocated (yet?).

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 6 of 14 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/lib.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDDart-AwKTLgkqdOu:-LDawgfb9P1vSkf3smct:bp1hwvy) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L14)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

This is more of a step-by-step guide, it's not related to `ping_service` example. I'll reword that part to be clearer.
</blockquote></details>

Are you happy with the new wording?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 6 of 14 files reviewed at latest revision, 3 unresolved discussions.

---

*[examples/ping_service.rs, line 104 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDD8SJC84OtEdjdhL5k:-LDazCCe8upHfihee5IX:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/examples/ping_service.rs#L104)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Alright man, but the best I can do is to move the localhost as array to static storage since `IpAddr` cannot be statically allocated (yet?).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Awesome addition of docs! Some fairly minor comments about that. But also a bigger concern about the actual thread safety of the current implementation.

---

Review status: 13 of 15 files reviewed at latest revision, 1 unresolved discussion.

---

*[examples/uninstall_service.rs, line 26 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuR-OU3_S9f0xo0feq:-LDuR-OU3_S9f0xo0fer:b-6uvunb) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/examples/uninstall_service.rs#L26)):*
> ```Rust
>     }
> 
>     service.delete().unwrap();
> ```

With the newly added support for returning `Result` from `main` I think examples like these can make use of `?` and thus be both shorter and encourage nicer error handling. I think that's one of the main reasons the feature was added.

The API guidelines has always mandated "proper error handling" in docs. But now it's finally possible to do without a lot of boilerplate.

https://rust-lang-nursery.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark

---

*[src/lib.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDDart-AwKTLgkqdOu:-LDuRxpn4TTOrPQ4zbu8:bttc1nq) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L14)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Are you happy with the new wording?
</blockquote></details>

Yes! I think this is really nice :ok_hand: 

---

*[src/lib.rs, line 76 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDDHXzXDTliyBoQVR4u:-LDuWwKd5V4O-lKC95CP:b-ph6fcu) ([raw file](https://github.com/mullvad/windows-service-rs/blob/e2f912e9c29f977d5b185cff7675a8ee53a6e3f4/src/lib.rs#L76)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Renamed to `register` for now.
</blockquote></details>

Good argument about register vs status handle. The rename to a shorter name is nice, we'll stick to that then I guess.

---

*[src/lib.rs, line 55 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuSvXx14r6cPJ3oiLg:-LDuSvXx14r6cPJ3oiLh:b-50tsos) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L55)):*
> ```Rust
> //!
> //! It's worth to mention that events are dispatched concurrently so it's important to make sure
> //! that your code is thread safe, the simplest way is to use [`std::sync::mpsc::channel`].
> ```

Great to mention the thread safety requirement. Is this expressed in Rust, so it's actually impossible to break that requirement without `unsafe`? It really should be since this is a safe high level wrapper. I think that maybe we need to add a `Send + Sync` bound to the closure passed in to prohibit people from passing closures that are not allowed to traverse into other threads or be called in parallel.

Also, `mpsc::Sender`s are not `Sync`. So I don't think it's safe to call them in parallel. If the closure can actually be invoked in parallel I think one would need to wrap it in a `Mutex`.

I think we should first figure out the bounds needed to actually make the `register` function safe. Then after that the compiler will tell us what we need to do in order to make the closures be accepted again.

---

*[src/lib.rs, line 91 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZ6eLEnlnX3PQ_dUM:-LDuZ6eLEnlnX3PQ_dUN:bcdy1j0) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L91)):*
> ```Rust
> //!
> //! The application needs to complete the initialization, obtain [`ServiceStatusHandle`] (see
> //! [`service_control_handler::register`]) and transition to [`Running`] state.
> ```

I can't find the corresponding link for this `[]` at the bottom of this crate documentation.

---

*[src/lib.rs, line 96 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZOGO67nUH3LdxDhG:-LDuZOGP73Cz0gRTcFS7:b5bk8zt) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L96)):*
> ```Rust
> //! much time it needs to complete it, by sending the [`StartPending`] state, time
> //! estimate using [`ServiceStatus::wait_hint`] and increment [`ServiceStatus::checkpoint`] each
> //! time the service completes the step in initialization.
> ```

completes **a** step?

---

*[src/lib.rs, line 98 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZWDi9mAGvbeJgnK-:-LDuZWDi9mAGvbeJgnK0:b-icmzue) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L98)):*
> ```Rust
> //! time the service completes the step in initialization.
> //!
> //! The system will attempt to kill the service that is not able to transition to [`Running`]
> ```

"kill **a** service"? Or alternatively "... to kill the service, if it is not able to ..."

---

*[src/sc_handle.rs, line 4 at r5](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDfoeZv-fgNqK0PIU35:-LDuNa1v-khv3qm-SORY:bb6mt9b) ([raw file](https://github.com/mullvad/windows-service-rs/blob/8f83927dfb43ba4e4fff5d323ac3b59b46ea6e1d/src/sc_handle.rs#L4)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

MS documentation references these as object handles in their documentation, basically either handles for services or service managers.

`ObjectHandle` was equally bad name so I decided to keep the original `ScHandle` which is somehow recognizable...

The point of this struct is to primarily get rid of handle management from both `Services` and `ServiceManagers`.

Secondarily, some of the advanced functionality, such as service state notifications require the valid handle. That means that we need to share the handle somehow since it may outlive the `Service` that previously used to `Drop` the handle. 

When the `Drop` is moved to `ScHandle` we can easily share it using `Arc<ScHandle>` and the handle will be properly released once it's not used anymore.
</blockquote></details>

:+1:

---

*[src/service.rs, line 378 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuQ9PE8AcLeqoqhltc:-LDuQ9PF-m7ZNiZDFloe:b-7oqr5s) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service.rs#L378)):*
> ```Rust
> /// The instances of the [`Service`] can be obtained via [`ServiceManager`].
> ///
> /// [`ServiceManager`]: super::service_manager::ServiceManager
> ```

Does this way of referencing documentation of other structs work? :open_mouth: That is awesome. I can't easily render the Windows docs here.

Not even the API guidelines mention this way of linking, so it it brand new?

---

*[src/service_control_handler.rs, line 80 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuUjYo36Ft9wUD4DuH:-LDuUjYo36Ft9wUD4DuI:bso9ncc) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service_control_handler.rs#L80)):*
> ```Rust
> ///
> /// Returns [`ServiceStatusHandle`] that can be used to report the service status back to the
> /// system.
> ```

The part about being called in parallel and having to be thread safe is mentioned in the crate docs, which is awesome. But it's not mentioned here. Might be worth adding that same text here as well?

---

*[src/service_control_handler.rs, line 102 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuU9RE2U3PpaU5BCyH:-LDuU9RE2U3PpaU5BCyI:b-ct54uj) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service_control_handler.rs#L102)):*
> ```Rust
> /// # fn main() {}
> /// ```
> pub fn register<S: AsRef<OsStr>, F: Fn(ServiceControl) -> ServiceControlHandlerResult + 'static>(
> ```

When the generics and their bounds become very long it's usually nicer to put the bounds in the `where` position. Up to you, just noting this is a fairly long one :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-:-LDuaY_E7fZWItwIiLHp:b-ygacyf)*

pronebird:
Review status: 13 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/lib.rs, line 55 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuSvXx14r6cPJ3oiLg:-LDunawA-pU-xXhfrND1:b-8sawby) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L55)):*

Great to mention the thread safety requirement. Is this expressed in Rust, so it's actually impossible to break that requirement without `unsafe`? It really should be since this is a safe high level wrapper. I think that maybe we need to add a `Send + Sync` bound to the closure passed in to prohibit people from passing closures that are not allowed to traverse into other threads or be called in parallel.

Also, `mpsc::Sender`s are not `Sync`. So I don't think it's safe to call them in parallel. If the closure can actually be invoked in parallel I think one would need to wrap it in a `Mutex`.

I think we should first figure out the bounds needed to actually make the `register` function safe. Then after that the compiler will tell us what we need to do in order to make the closures be accepted again.
</blockquote></details>

From my experience the event handler is called on one of background threads, so maybe I should clarify that. It's not called on each thread each time. I can verify that btw, but last time I printed the thread id it matched across calls.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 13 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/lib.rs, line 91 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZ6eLEnlnX3PQ_dUM:-LDuoNC05sgsb1lOawvs:b-wndi3n) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L91)):*

I can't find the corresponding link for this `[]` at the bottom of this crate documentation.
</blockquote></details>

Run `cargo doc` and you'll see it automatically resolves this. I'll boot to windows shortly to double check that.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: 13 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/lib.rs, line 55 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuSvXx14r6cPJ3oiLg:-LDupGxgCLpE-XXNPOwh:bgr9g1z) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L55)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

From my experience the event handler is called on one of background threads, so maybe I should clarify that. It's not called on each thread each time. I can verify that btw, but last time I printed the thread id it matched across calls.
</blockquote></details>

So it's never called from multiple threads simultaneously (in parallel)? Then it does not need to be thread safe?

I still think we might need the `Send` bound on the closure though. In order to only allow closures that can be passed between threads.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 13 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/lib.rs, line 55 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuSvXx14r6cPJ3oiLg:-LDup6xfBuI3al1ty2Qb:b6ytehe) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L55)):*

So it's never called from multiple threads simultaneously (in parallel)? Then it does not need to be thread safe?

I still think we might need the `Send` bound on the closure though. In order to only allow closures that can be passed between threads.
</blockquote></details>

Sometimes I rush and juggle many things at once and then I leave a lot of typos in my replies. 

What I meant is that event handler is called on a single background thread. So there is no need to wrap `Sender` into `Mutex`.

I'll verify that just in case. But I believe it would be impossible for Windows service to call event handler concurrently since it actually waits for response from the event handler.

I agree that we need to add `Send`. I was not sure there was a way to tell Rust that `Fn` is going to be dispatched on some thread behind the scenes (i.e. from inside FFI function)

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 13 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/lib.rs, line 96 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZOGO67nUH3LdxDhG:-LDussbtCh8xony-Qimx:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L96)):*

completes **a** step?
</blockquote></details>

Done.

---

*[src/lib.rs, line 98 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZWDi9mAGvbeJgnK-:-LDusssq-_53HGl4f-pd:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L98)):*

"kill **a** service"? Or alternatively "... to kill the service, if it is not able to ..."
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 12 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/service.rs, line 378 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuQ9PE8AcLeqoqhltc:-LDut5AW2-u9PUjbSEfs:b25gv1x) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service.rs#L378)):*

Does this way of referencing documentation of other structs work? :open_mouth: That is awesome. I can't easily render the Windows docs here.

Not even the API guidelines mention this way of linking, so it it brand new?
</blockquote></details>

I'll double check but I think I already checked that with `cargo doc` and it worked fine.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 11 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/service_control_handler.rs, line 102 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuU9RE2U3PpaU5BCyH:-LDuvBuABqolDcmiGtmj:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service_control_handler.rs#L102)):*

When the generics and their bounds become very long it's usually nicer to put the bounds in the `where` position. Up to you, just noting this is a fairly long one :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 11 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/service_control_handler.rs, line 80 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuUjYo36Ft9wUD4DuH:-LDuvROQ6j2XBJrGDGtb:bcp2sb7) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service_control_handler.rs#L80)):*

The part about being called in parallel and having to be thread safe is mentioned in the crate docs, which is awesome. But it's not mentioned here. Might be worth adding that same text here as well?
</blockquote></details>

`ServiceStatusHandle` implements `Send`. Is there any need to re-iterate that?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: 11 of 15 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/service_control_handler.rs, line 80 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuUjYo36Ft9wUD4DuH:-LDuxvTN0t_vB6-TtzLe:b-xxeqyz) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service_control_handler.rs#L80)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

`ServiceStatusHandle` implements `Send`. Is there any need to re-iterate that?
</blockquote></details>

Hmm. That is a completely different thing, no? I'm talking about `event_handler` needing to be thread safe. But this is the same thing as the other discussion, so let's keep the discussion about the properties of that over there until we have arrived at a conclusion.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: 11 of 15 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/lib.rs, line 55 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuSvXx14r6cPJ3oiLg:-LDuyB166t-z4RLrvAIi:b-kn2lnv) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L55)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Sometimes I rush and juggle many things at once and then I leave a lot of typos in my replies. 

What I meant is that event handler is called on a single background thread. So there is no need to wrap `Sender` into `Mutex`.

I'll verify that just in case. But I believe it would be impossible for Windows service to call event handler concurrently since it actually waits for response from the event handler.

I agree that we need to add `Send`. I was not sure there was a way to tell Rust that `Fn` is going to be dispatched on some thread behind the scenes (i.e. from inside FFI function)
</blockquote></details>

If it is only ever called from a single thread it does indeed not need to be thread safe (`Sync`). But yeah, since it will be passed to, and called from, another thread than where the closure originates it must be allowed to be sent between threads (`Send`).

If that is true I don't think we should mention thread safety nor concurrency. It's basically just transferring the closure to a system thread and then it stays there, being called in a non-overlapping fashion.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 11 of 15 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/lib.rs, line 91 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZ6eLEnlnX3PQ_dUM:-LDuyqNk9E-e0Hm9wQ_V:b-f9e00w) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L91)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Run `cargo doc` and you'll see it automatically resolves this. I'll boot to windows shortly to double check that.
</blockquote></details>

Double checked, it works as expected. Probably new addition to resolver.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 11 of 15 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/lib.rs, line 91 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuZ6eLEnlnX3PQ_dUM:-LDuzFHE-vdoAOYsEZLo:b-e0gyja) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L91)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Double checked, it works as expected. Probably new addition to resolver.
</blockquote></details>



---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: 14 of 15 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 14 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 55 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuSvXx14r6cPJ3oiLg:-LDv8W7_3M2epq-s_16S:b-jo04pj) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L55)):*

If it is only ever called from a single thread it does indeed not need to be thread safe (`Sync`). But yeah, since it will be passed to, and called from, another thread than where the closure originates it must be allowed to be sent between threads (`Send`).

If that is true I don't think we should mention thread safety nor concurrency. It's basically just transferring the closure to a system thread and then it stays there, being called in a non-overlapping fashion.
</blockquote></details>

I ran a series of calls and logged the thread name and ID:

```
main(): thread ThreadId(0), Some("main")
my_service_main: thread ThreadId(1), None
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
```

Seems like calls to event_handler are in fact dispatched on a single thread and actually the same where we start the dispatcher, which makes me think that `service_dispatcher::start` simply polls events.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 14 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/service_control_handler.rs, line 80 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuUjYo36Ft9wUD4DuH:-LDvA_xiAetk7FQoUcLM:b-z3agay) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/service_control_handler.rs#L80)):*

Hmm. That is a completely different thing, no? I'm talking about `event_handler` needing to be thread safe. But this is the same thing as the other discussion, so let's keep the discussion about the properties of that over there until we have arrived at a conclusion.
</blockquote></details>

OK, I thought you pointed the other line.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 13 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 55 at r6](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDuSvXx14r6cPJ3oiLg:-LDvCIxjBlm7Xx2W3GbL:ba2mupy) ([raw file](https://github.com/mullvad/windows-service-rs/blob/91b650144c1c96c211a948c7f4fb66503d7a5b1d/src/lib.rs#L55)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

I ran a series of calls and logged the thread name and ID:

```
main(): thread ThreadId(0), Some("main")
my_service_main: thread ThreadId(1), None
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
event_handler: thread ThreadId(0), Some("main")
```

Seems like calls to event_handler are in fact dispatched on a single thread and actually the same where we start the dispatcher, which makes me think that `service_dispatcher::start` simply polls events.
</blockquote></details>

I removed the mention about concurrency and updated README.md.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: 15 of 16 files reviewed at latest revision, all discussions resolved.

---

*[appveyor.yml, line 5 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNe2L2nPYu5z3_-jL:-LDvNe2M4bfVZsRQeKMn:b-qsmh66) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L5)):*
> ```YAML
> 
> image:
> - Visual Studio 2017
> ```

The AppVeyor config from `mullvadvpn-app` has a lot of stuff we don't really need for pure Rust. It has been added because of our C++ dependencies.

I think we can take inspiration from the original source (trust) as the comment at the top says. We probably don't need the `image` nor `platform` sections.

---

*[appveyor.yml, line 60 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNuRz2SsiMnvkpc1_:-LDvNuRz2SsiMnvkpc1a:b-42uwii) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L60)):*
> ```YAML
> # We build in the test stage, so disable Appveyor's build stage. This prevents
> # the "directory does not contain a project or solution file" error.
> build: false
> ```

This seems to be a funky ending. Do you maybe have an `\r` here or something?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 15 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[appveyor.yml, line 5 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNe2L2nPYu5z3_-jL:-LDvPzx71kNB2pExc1TB:b-ieprjx) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L5)):*

The AppVeyor config from `mullvadvpn-app` has a lot of stuff we don't really need for pure Rust. It has been added because of our C++ dependencies.

I think we can take inspiration from the original source (trust) as the comment at the top says. We probably don't need the `image` nor `platform` sections.
</blockquote></details>

Removed irrelevant sections

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 15 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[appveyor.yml, line 60 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNuRz2SsiMnvkpc1_:-LDvQ5qZDxyMAnaNASdk:biv2c1t) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L60)):*

This seems to be a funky ending. Do you maybe have an `\r` here or something?
</blockquote></details>

It means there is no newline.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 15 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[appveyor.yml, line 60 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNuRz2SsiMnvkpc1_:-LDvQCUW2SoZ92LAnw8x:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L60)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

It means there is no newline.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:
Review status: 15 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[appveyor.yml, line 5 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNe2L2nPYu5z3_-jL:-LDvQEFM9KYB4ZYI6rIT:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L5)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Removed irrelevant sections
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[appveyor.yml, line 60 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNuRz2SsiMnvkpc1_:-LDvRSuOCKppZC7BLyWK:bttodsw) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L60)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

The arrow yes. But it also had an `^M` byte in it. Anyway, gone now.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

faern:
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[appveyor.yml, line 6 at r11](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvSytP9NvFllTfbPee:-LDvSytP9NvFllTfbPef:bmrmrem) ([raw file](https://github.com/mullvad/windows-service-rs/blob/7190715c23b880657943dc3aae5ec43c0d989980/appveyor.yml#L6)):*
> ```YAML
> environment:
>   global:
>     CRATE_NAME: "windows-service"
> ```

The `trust` example config does not have quotes. But it seems to work with as well so it doesn't matter I guess.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:

---

*[appveyor.yml, line 6 at r11](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvSytP9NvFllTfbPee:-LDvTU6l9Rk4KxkxTBqy:bks44gs) ([raw file](https://github.com/mullvad/windows-service-rs/blob/7190715c23b880657943dc3aae5ec43c0d989980/appveyor.yml#L6)):*

The `trust` example config does not have quotes. But it seems to work with as well so it doesn't matter I guess.
</blockquote></details>

I pushed a change with quotes (realized later)

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

pronebird:

---

*[appveyor.yml, line 60 at r9](https://reviewable.io/reviews/mullvad/windows-service-rs/1#-LDvNuRz2SsiMnvkpc1_:-LDvT_kJ3H5XJAfwSOVF:b2nj6od) ([raw file](https://github.com/mullvad/windows-service-rs/blob/1e8c833c1daf0680047251dcbd5620ab862081a3/appveyor.yml#L60)):*

The arrow yes. But it also had an `^M` byte in it. Anyway, gone now.
</blockquote></details>

My bad. It defaulted to CRLF.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/1)*

PR #3, opened 2018-06-03
Unify errors and handle errors in examples

faern:
At first I noticed that we had many different error types, and all had the `System(::std::io::Error) #[doc = "System call error."];` variant in it. Many of them also had other overlapping/equal error variants.

`error-chain` itself mandates having one crate-global error that the library exports. We have previously not followed this because we usually have very different errors in different modules, so it does not make much sense to group them etc. But here I felt they did overlap and it made a lot of sense to have one common `windows_service::Error`.

Then I also wanted to play around with the new functionality that `main` can return a `Result`. So I thought I'll try to incorporate that into the documentation in order to do "proper" error handling in the examples and documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 name = "windows-service"
 version = "0.1.0"
 description = "A crate that provides facilities for management and implementation of windows services"
+readme = "README.md"
 authors = ["Mullvad VPN <admin@mullvad.net>", "Andrej Mihajlov <and@mullvad.net>", "Linus Färnstrand <linus@mullvad.net>"]
 keywords = ["windows", "service", "daemon"]
 categories = ["api-bindings"]
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ fn run_service(arguments: Vec<OsString>) -> windows_service::Result<()> {
     };
 
     // Tell the system that the service is running now
-    status_handle.set_service_status(next_status);
+    status_handle.set_service_status(next_status)?;
 
     // Do some work
 
--- a/README.md
+++ b/README.md
@@ -33,10 +33,11 @@ fn my_service_main(arguments: Vec<OsString>) {
     // `service_dispatcher::start` from `main`.
 }
 
-fn main() {
+fn main() -> Result<(), windows_service::Error> {
     // Register generated `ffi_service_main` with the system and start the service, blocking
     // this thread until the service is stopped.
-    service_dispatcher::start("myservice", ffi_service_main).unwrap();
+    service_dispatcher::start("myservice", ffi_service_main)?;
+    Ok(())
 }
 ```
 
@@ -53,6 +54,12 @@ use windows_service::service::ServiceControl;
 use windows_service::service_control_handler::{self, ServiceControlHandlerResult};
 
 fn my_service_main(arguments: Vec<OsString>) {
+    if let Err(_e) = run_service(arguments) {
+        // Handle errors in some way.
+    }
+}
+
+fn run_service(arguments: Vec<OsString>) -> Result<(), windows_service::Error> {
     let event_handler = move |control_event| -> ServiceControlHandlerResult {
         match control_event {
             ServiceControl::Stop => {
@@ -66,7 +73,8 @@ fn my_service_main(arguments: Vec<OsString>) {
     };
 
     // Register system service event handler
-    let status_handle = service_control_handler::register("myservice", event_handler).unwrap();
+    let status_handle = service_control_handler::register("myservice", event_handler)?;
+    Ok(())
 }
 ```
 
@@ -106,6 +114,12 @@ use windows_service::service::{
 use windows_service::service_control_handler::{self, ServiceControlHandlerResult};
 
 fn my_service_main(arguments: Vec<OsString>) {
+    if let Err(_e) = run_service(arguments) {
+        // Handle error in some way.
+    }
+}
+
+fn run_service(arguments: Vec<OsString>) -> windows_service::Result<()> {
     let event_handler = move |control_event| -> ServiceControlHandlerResult {
         match control_event {
             ServiceControl::Stop | ServiceControl::Interrogate => {
@@ -116,8 +130,7 @@ fn my_service_main(arguments: Vec<OsString>) {
     };
 
     // Register system service event handler
-    let status_handle =
-        service_control_handler::register("my_service_name", event_handler).unwrap();
+    let status_handle = service_control_handler::register("my_service_name", event_handler)?;
 
     let next_status = ServiceStatus {
         // Should match the one from system service registry
@@ -138,8 +151,9 @@ fn my_service_main(arguments: Vec<OsString>) {
     status_handle.set_service_status(next_status);
 
     // Do some work
-}
 
+    Ok(())
+}
 ```
 
 Please refer to the "Service State Transitions" article on MSDN for more info:\
--- a/examples/install_service.rs
+++ b/examples/install_service.rs
@@ -2,7 +2,7 @@
 extern crate windows_service;
 
 #[cfg(windows)]
-fn main() {
+fn main() -> windows_service::Result<()> {
     use std::ffi::OsString;
     use windows_service::service::{
         ServiceAccess, ServiceErrorControl, ServiceInfo, ServiceStartType, ServiceType,
@@ -10,7 +10,7 @@ fn main() {
     use windows_service::service_manager::{ServiceManager, ServiceManagerAccess};
 
     let manager_access = ServiceManagerAccess::CONNECT | ServiceManagerAccess::CREATE_SERVICE;
-    let service_manager = ServiceManager::local_computer(None::<&str>, manager_access).unwrap();
+    let service_manager = ServiceManager::local_computer(None::<&str>, manager_access)?;
 
     // This example installs the service defined in `examples/ping_service.rs`.
     // In the real world code you would set the executable path to point to your own binary
@@ -30,9 +30,8 @@ fn main() {
         account_name: None, // run as System
         account_password: None,
     };
-    let _service = service_manager
-        .create_service(service_info, ServiceAccess::empty())
-        .unwrap();
+    let _service = service_manager.create_service(service_info, ServiceAccess::empty())?;
+    Ok(())
 }
 
 #[cfg(not(windows))]
--- a/examples/ping_service.rs
+++ b/examples/ping_service.rs
@@ -17,8 +17,8 @@
 extern crate windows_service;
 
 #[cfg(windows)]
-fn main() {
-    ping_service::run();
+fn main() -> windows_service::Result<()> {
+    ping_service::run()
 }
 
 #[cfg(not(windows))]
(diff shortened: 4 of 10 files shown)


pronebird:
Review status: 3 of 10 files reviewed at latest revision, all discussions resolved.

---

*[examples/uninstall_service.rs, line 24 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9DbBWBSXUlxsbVwpU:-LE9DbBWBSXUlxsbVwpV:b-v82hys) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/examples/uninstall_service.rs#L24)):*
> ```Rust
>     }
> 
>     service.delete()?;
> ```

Could probably end with `service.delete()` and that's it. 

However, `service.delete()` returns `io::Result`, so I am not sure this will compile without `into()`

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

pronebird:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/lib.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9M8c9EkLyyo_2dy9B:-LE9M8cAESjdR9pS9aFa:b-1a54fe) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L157)):*
> ```Rust
> //!
> //!     // Tell the system that the service is running now
> //!     status_handle.set_service_status(next_status);
> ```

Could be my bad, but `set_service_status` returns `io::Result`

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

pronebird:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 177 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9N-gY5_9nQeButkSC:-LE9N-gY5_9nQeButkSD:bt9adx) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L177)):*
> ```Rust
> 
> // Because of how deeply error-chain recurse with this many error types.
> #![recursion_limit = "128"]
> ```

I have no idea how error chain expands its macro, but it seems like we have a flat enum of errors so where does such long recursion come from?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

pronebird:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/lib.rs, line 63 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9PFXPBvq_BvpMQBNE:-LE9PFXPBvq_BvpMQBNF:b1490uk) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L63)):*
> ```Rust
> //!
> //! fn my_service_main(arguments: Vec<OsString>) {
> //!     if let Err(_e) = run_service(arguments) {
> ```

While I see where it comes from, I think it adds a slight indirection to example. I believe that examples are the best when they are as simple as possible, shorter - better.

IMO examples are not meant to be copy and pasted but rather show how to use something with a happy code path.

It's up to developer to introduce proper error handling and cover all edge cases.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

faern:
Review status: 8 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[examples/uninstall_service.rs, line 24 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9DbBWBSXUlxsbVwpU:-LE9Q2tm0aAL-XQG2oRV:b7g31la) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/examples/uninstall_service.rs#L24)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Could probably end with `service.delete()` and that's it. 

However, `service.delete()` returns `io::Result`, so I am not sure this will compile without `into()`
</blockquote></details>

Yes, one reason for `?` is to get the `.into()` automatically. Another is to make it clear and easy to see that this is a fallible operation that requires error handling. I think fallible methods that just end with a fallible expression can be kind of vague. Especially in an example it's good to fully try to show what's going on. Which is why I prefer handling all errors and then having a separate `Ok(())` to return with.

---

*[src/lib.rs, line 63 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9PFXPBvq_BvpMQBNE:-LE9QluL0kJXhiSfPiLg:b-hw459w) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L63)):*
The API guidelines say exactly the opposite. https://rust-lang-nursery.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark

> Like it or not, example code is often copied verbatim by users. Unwrapping an error should be a conscious decision that the user needs to make.

And I also personally like it when examples are so good that you can just copy and paste them to get started, and then iterate from there.

---

*[src/lib.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9M8c9EkLyyo_2dy9B:-LE9RS8E533Sl2OlSjAn:b2h3oj3) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L157)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

Could be my bad, but `set_service_status` returns `io::Result`
</blockquote></details>

Fixed :ok_hand: 

---

*[src/lib.rs, line 177 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9N-gY5_9nQeButkSC:-LE9RVw_CXF3X7g0NoyQ:btjwws0) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L177)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

I have no idea how error chain expands its macro, but it seems like we have a flat enum of errors so where does such long recursion come from?
</blockquote></details>

The result of the macro is a flat enum indeed. And I have not read the macro myself in a very long time. But I guess it recurses a number of times for each variant it generates. I'm not sure.

This recursion limit is only for the compile process, it does not affect the runtime in any way.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

faern:
Review status: 8 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/lib.rs, line 63 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9PFXPBvq_BvpMQBNE:-LE9Rw1d3hy8s5FrA2ot:b-qmc7e7) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L63)):*

The API guidelines say exactly the opposite. https://rust-lang-nursery.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark

> Like it or not, example code is often copied verbatim by users. Unwrapping an error should be a conscious decision that the user needs to make.

And I also personally like it when examples are so good that you can just copy and paste them to get started, and then iterate from there.
</blockquote></details>

I don't think it adds very much complexity to the example. And the benefit is that it presents a reasonable way to handle errors. Something the user might spend time on figuring out how to do otherwise.

---

*[src/lib.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9M8c9EkLyyo_2dy9B:-LE9RsVlBeTPfoFuZwip:b-896fix) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L157)):*

Fixed :ok_hand: 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

pronebird:

---

*[src/lib.rs, line 63 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9PFXPBvq_BvpMQBNE:-LE9eMtACoScCmc_l9BF:b-vqsugi) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L63)):*

I don't think it adds very much complexity to the example. And the benefit is that it presents a reasonable way to handle errors. Something the user might spend time on figuring out how to do otherwise.
</blockquote></details>

This example only features a single aspect of the windows service programming, so copying it as-is gives very little benefit. 

The complete example is likely to be arranged differently, partially because runtime or initialization errors in windows service are better be reported along with the `Stopped` status.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

faern:

---

*[src/lib.rs, line 63 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9PFXPBvq_BvpMQBNE:-LE9gTNh33N45CgsVVsc:b-1h6ato) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L63)):*
<details><summary><i>Previously, pronebird (Andrei Mihailov) wrote…</i></summary><blockquote>

This example only features a single aspect of the windows service programming, so copying it as-is gives very little benefit. 

The complete example is likely to be arranged differently, partially because runtime or initialization errors in windows service are better be reported along with the `Stopped` status.
</blockquote></details>

Yes, the `event_handler` and `service_control_handler::register` is indeed the parts that are being showcased in this example. And part if showing it is showing where/when it should be called, which is why `my_service_main` has to be present(?).

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

pronebird:

---

*[src/lib.rs, line 63 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-LE9PFXPBvq_BvpMQBNE:-LE9hXE9E0g05UCVq-Io:b56myuj) ([raw file](https://github.com/mullvad/windows-service-rs/blob/abd8474729361ac6915176f7fbe94f4ddfe1f2b8/src/lib.rs#L63)):*

Yes, the `event_handler` and `service_control_handler::register` is indeed the parts that are being showcased in this example. And part if showing it is showing where/when it should be called, which is why `my_service_main` has to be present(?).
</blockquote></details>

Yeah, it shows the connection between two and when to call `::register`

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3)*

pronebird:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/3#-:-LE9krciDijQTP9XvE6-:bnfp4nl)*

PR #19, opened 2018-11-09
Bump 0.2.0

pronebird:
We should release an update at some point. I've bumped the version to 0.2.0 because the recent change to `ServiceType` breaks the backward compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/19)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,21 +5,27 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+
+## [0.2.0] - 2019-04-01
 ### Added
 - Add support for specifying service dependencies when creating a service.
 - A `ServiceExitCode::NO_ERROR` constant for easy access to the success value.
 - Add `Service::start` for starting services programmatically.
+- Add `Service::query_config` for getting the config of the service.
 
 ### Changed
 - Changed `service_control_handler::register` to accept an `FnMut` rather than just an `Fn` for the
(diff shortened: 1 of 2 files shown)


jheander:
Would be awesome if this release could be published. Right now the documented example on the Github readme does not compile since the ServiceType name does not match the version distributed over cargo.

faern:
The documentation for a release is available under that release. The README corresponding to the release is published on crates.io: https://crates.io/crates/windows-service

We are in the process of getting rid of `error-chain` here, and that is also a breaking change. So I think it would make a lot of sense to fix that before publishing `0.2.0`, otherwise we will need to release `0.3.0` very soon again.

jheander:
Ah, that makes sense. Technically since the changes are not backwards compatible I guess both changes should really affect the major version, so that the next release becomes `1.0.0`.

pronebird:
@jheander AFAIK the versions that start with the major zero are treated differently in semver, so 0.2 and 0.3 would be treated the same as 2.0 and 3.0 - more like a major bump than minor.

faern:
Exactly. Semver is a bit special for versions `0.x.y`. There `x` is treated as the major version.

AviateX14:
Do you have a timeline for delivering 0.2.0 with the error-chain changes? The service status querying doesn't seem to work in 0.1.0 as far as I can see, but pulling directly from git, everything works fine,

faern:
We will publish it real soon. We just wanted to test it a bit after merging the last few changes. But if you also report it works fine from the master branch, then we might just as well publish it now.

faern:
@AviateX14 @jheander This has now been merged. And 0.2.0 has been published to crates.io along with a signed git tag uploaded to this git repository.

PR #22, opened 2019-01-25
feature toggle backtrace in error-chain for legacy os

rustysec:
`error-chain` conditionally uses `backtrace` under the hood which requires a call to `SymFromAddrW`. This is part of `DbgHelp.dll` which is not supported on older operating systems such as XP or Server 2003. Currently, using `windows-service-rs` causes an executable to crash at start because of this missing symbol. This PR introduces a way to feature flag `backtrace` off in `error-chain` without changing anything else in the library. I've tested this against Windows 10 and Server 2003 using the examples and it all seems to be functioning as intended.

`Cargo.toml` usage:
```
windows-service = { version = "*", default-features = false, features = ["legacy-support"] }
```

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/22)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,12 @@ edition = "2018"
 
 [target.'cfg(windows)'.dependencies]
 bitflags = "1.0.1"
-error-chain = "0.12"
+error-chain = { version = "0.12", optional = true }
+error-chain-without-backtrace = { version = "0.12", package = "error-chain", default-features = false, optional = true }
 winapi = { version = "0.3.5", features = ["std", "winsvc", "winerror"] }
 widestring = "0.3.0"
+
+[features]
+default = ["with-backtrace"]
+with-backtrace = ["error-chain"]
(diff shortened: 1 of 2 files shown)


rustysec:
well, unfortunately my method doesn't seem to play nice when other deps bring in `error-chain` without the rename. ```multiple dependencies listed for the same crate must all have the same name, but the dependency on `error-chain vX.Y.Z` is listed as having different names``` from cargo. i'll re-open a PR when I have a better solution.

pronebird:
@rustysec I think someone else already tried to solve this before https://github.com/mullvad/windows-service-rs/pull/7

It can be that we will be moving away from the `error_chain` crate at some point, but no ETA on that yet.

rustysec:
well earlier today i moved this crate over to failure just for fun. i can open a pr for that if you'd like. i'm struggling to track down some other dependencies of mine that are posing the same problem. all kinds of chaos when you have to support legacy systems.

PR #24, opened 2019-03-14
implement failure over error-chain

rustysec:
i finally got around to taking this on. all the tests are passing, but i haven't tested it fully. wanted to get some eyes on this and see if it's still something you are interested in pursuing.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/24)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,7 +39,7 @@
 //!     // `service_dispatcher::start` from `main`.
 //! }
 //!
-//! fn main() -> Result<(), windows_service::Error> {
+//! fn main() -> Result<()> {
 //!     // Register generated `ffi_service_main` with the system and start the service, blocking
 //!     // this thread until the service is stopped.
 //!     service_dispatcher::start("myservice", ffi_service_main)?;
@@ -65,7 +65,7 @@
 //!     }
 //! }
 //!
-//! fn run_service(arguments: Vec<OsString>) -> Result<(), windows_service::Error> {
+//! fn run_service(arguments: Vec<OsString>) -> Result<()> {
 //!     let event_handler = move |control_event| -> ServiceControlHandlerResult {
 //!         match control_event {
 //!             ServiceControl::Stop => {
@@ -175,10 +175,10 @@
 
 #[macro_use]
 extern crate err_derive;
-pub type Result<T> = std::result::Result<T, ErrorKind>;
+pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Error, Debug)]
-pub enum ErrorKind {
+pub enum Error {
     /// Invalid account name.
     #[error(display = "Invalid account name")]
     InvalidAccountName,
@@ -208,7 +208,7 @@ pub enum ErrorKind {
     InvalidServiceName,
     /// Invalid start argument.
     #[error(display = "Invalid start argument")]
-    InvalidStartArgument,
+    InvalidStartArgument(#[error(cause)] widestring::NulError),
     /// Invalid raw representation of [`ServiceState`].
     #[error(display = "Invalid service state value: {}", _0)]
     InvalidServiceState(u32),
@@ -222,13 +222,13 @@ pub enum ErrorKind {
     #[error(display = "Invalid service error control type: {}", _0)]
     InvalidServiceErrorControl(u32),
     /// IO error
-    #[error(display = "IO Error: {}", _0)]
-    IOError(String),
+    #[error(display = "IO Error")]
+    IOError(#[error(cause)] std::io::Error),
 }
 
-impl From<std::io::Error> for ErrorKind {
+impl From<std::io::Error> for Error {
     fn from(error: std::io::Error) -> Self {
-        ErrorKind::IOError(error.to_string())
+        Error::IOError(error)
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,7 +30,7 @@
 //! extern crate windows_service;
 //!
 //! use std::ffi::OsString;
-//! use windows_service::service_dispatcher;
+//! use windows_service::{Result, service_dispatcher};
 //!
 //! define_windows_service!(ffi_service_main, my_service_main);
 //!
@@ -58,6 +58,7 @@
 //! use std::ffi::OsString;
 //! use windows_service::service::ServiceControl;
 //! use windows_service::service_control_handler::{self, ServiceControlHandlerResult};
+//! use windows_service::Result;
 //!
 //! fn my_service_main(arguments: Vec<OsString>) {
 //!     if let Err(_e) = run_service(arguments) {
@@ -235,6 +236,10 @@ pub enum Error {
     #[error(display = "Invalid service error control type: {}", _0)]
     InvalidServiceErrorControl(u32),
 
+    /// A generic IO error
+    #[error(display = "An IO error has been encountered")]
+    IoError(#[error(cause)] std::io::Error),
+
     /// Failed to send command to service Service deletion to start
     #[error(display = "Could not send commands to service")]
     ServiceControlFailed(#[error(cause)] std::io::Error),
@@ -266,6 +271,16 @@ pub enum Error {
     /// Service failed to start
     #[error(display = "Could not start service")]
     ServiceStartFailed(#[error(cause)] std::io::Error),
+
+    /// Failed to set service status
+    #[error(display = "Could not set service status")]
+    ServiceStatusFailed(#[error(cause)] std::io::Error),
+}
+
+impl From<std::io::Error> for Error {
+    fn from(error: std::io::Error) -> Self {
+        Error::IoError(error)
+    }
 }
 
 mod sc_handle;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -172,82 +172,56 @@
 //! [`Running`]: service::ServiceState::Running
 
 #![cfg(windows)]
-// Because of how deeply error-chain recurse with this many error types.
-#![recursion_limit = "128"]
 
 #[macro_use]
-extern crate error_chain;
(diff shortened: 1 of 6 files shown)


src/service.rs
```diff
@@ -524,13 +523,26 @@ impl Service {
...
-        let mut raw_service_arguments: Vec<*const u16> =
-            wide_service_arguments.iter().map(|s| s.as_ptr()).collect();
+            /*
+            .map(|s| match WideCString::from_str(s) {
+                Err(_) => Err(ErrorKind::InvalidStartArgument.into()),
+                Ok(s) => Ok(s),
+            })
+            */
+            .map(|s| WideCString::from_str(s).map_err(|e| e.into()))
+            .collect::<Vec<Result<WideCString, Error>>>();
```
pronebird: Isn't it more convenient to collect into `Result<Vec<WideCString>, Error>` the way we used to?

Disclaimer: Haven't done rust in a while.

rustysec: i cleaned this up a bit, slightly more verbose than with `error-chain`

rustysec: changed the collection pattern to more closely resemble what it was with `error-chain`

src/lib.rs
```diff
@@ -172,82 +172,57 @@
 //! [`Running`]: service::ServiceState::Running
 
 #![cfg(windows)]
-// Because of how deeply error-chain recurse with this many error types.
-#![recursion_limit = "128"]
 
 #[macro_use]
-extern crate error_chain;
+extern crate failure;
+pub type Result<T> = std::result::Result<T, failure::Error>;
```
faern: I'm still trying to wrap my head around how libraries and applications should best do error management. But with my current view on it I don't think libraries should ever return the `failure::Error` type. It becomes quite opaque to the library consumer, it forces allocation and dynamic dispatch upon them, and the returned error does not implement the `std::error::Error` trait. I kind of like how `failure` propose we should manage errors. But I don't like how it potentially divides the ecosystem into those who implement `std::error::Error` and those who implement `failure::Fail`.

As such I think applications can go full on with `failure`, but I think libraries should try to conform to the standard library style and return custom specific errors. I have recently started using the `err-derive` crate for libraries. It's the exact same thing as `#[derive(Fail)]` but it implements the `std::error::Error` trait instead. See here for an example: https://github.com/mullvad/hyper-dnscache/blob/master/src/cache_storer.rs#L132-L144

That library does only help with the creation of the custom error types. It does not include anything related to throwing errors or collecting backtraces etc. But it's still quite convenient IMO. Here is an example: https://github.com/mullvad/hyper-dnscache/blob/master/src/cache_storer.rs#L31-L33

As I said, this is something I'm still figuring out how to do best. So please give your feedback on what you think about this.

src/service.rs
```diff
@@ -524,13 +523,18 @@ impl Service {
...
         let wide_service_arguments = service_arguments
             .iter()
-            .map(|s| WideCString::from_str(s).chain_err(|| ErrorKind::InvalidStartArgument))
-            .collect::<Result<Vec<WideCString>>>()?;
-        let mut raw_service_arguments: Vec<*const u16> =
-            wide_service_arguments.iter().map(|s| s.as_ptr()).collect();
+            .map(|s| WideCString::from_str(s).map_err(|_| ErrorKind::InvalidStartArgument))
+            .collect::<Vec<Result<WideCString, _>>>()
+            .into_iter()
+            .collect::<Result<Vec<WideCString>, _>>()?;
```
faern: Do you really need to `collect` twice? I think the magic of how `collect` works should be able to transform it into a `Result<Vec<...>, _>` directly?

rustysec: yeah that was unnecessary, my fault for not getting rid of it earlier.

src/lib.rs
rustysec: I wouldn't say I have a well developed opinion on that matter. A while back we were working a couple PRs to get better legacy system support and it was mentioned switching to `failure` was something the project might be interested in. I won't be heart broken if that opinion has changed.

faern: I'm aware of the previous discussions. And I'm sorry I might have changed opinion since then. That was before I read the `failure` guide (https://rust-lang-nursery.github.io/failure/use-error.html). After doing so I was less convinced `failure` is a good fit for libraries. And especially the `Error` struct. The guide mentions the downsides I talk about and say that one use case for that struct is prototyping, which is certainly not what we do here now :)

I think that the goal of `failure` is to fix the standard library error handling. But until `std::error::Error` fully becomes the `Fail` trait I see it more as an ecosystem breakage. This is where `err-derive` comes in and picks out some of the best parts of `failure` but makes it std compliant.

rustysec: No need to be sorry, things change! I'm fine with closing this pr out and just count it as a learning experience.

faern: No need to close it really! We still want to get rid of `error-chain`. And using `err-derive` to define the error types is almost identical to how you do it for `failure`. Check their documentation. So if you have the energy you could look into migrating the PR to do that with fairly few changes I think. If not, I still want to thank you for the effort and I will look into this myself when I have time later.

src/lib.rs
```diff
@@ -172,81 +172,63 @@
...
-// Because of how deeply error-chain recurse with this many error types.
-#![recursion_limit = "128"]
 
 #[macro_use]
-extern crate error_chain;
+extern crate err_derive;
+pub type Result<T> = std::result::Result<T, ErrorKind>;
 
+#[derive(Error, Debug)]
+pub enum ErrorKind {
```
faern: Please rename to `Error` now when it's an error (with causes and everything), and not just a kind.

src/lib.rs
```diff
@@ -172,81 +172,63 @@
...
-        InvalidServiceErrorControl(raw_value: u32) {
-            description("Invalid service error control type")
-            display("Invalid service error control type: {}", raw_value)
-        }
-    }
-    foreign_links {
-        System(::std::io::Error) #[doc = "System call error"];
+impl From<std::io::Error> for ErrorKind {
+    fn from(error: std::io::Error) -> Self {
+        ErrorKind::IOError(error.to_string())
```
faern: It's probably more appropriate to use the chaining functionality of `Error::source`. This can easily be done with `#[error(cause)]`. See https://github.com/mullvad/hyper-dnscache/blob/master/src/cache_storer.rs#L135 and https://github.com/mullvad/hyper-dnscache/blob/master/src/cache_storer.rs#L31

According to naming convention it should be spelled as `IoError`. However, depending on how it's used, it would probably be nicer for the user if it's split up into multiple errors, and not all errors that happens to originate from an `io::Error` lumped into one. Again, see the links above and compare `WriteFileError` with `ReadFileError`. This allows the error on this level to have a message telling what went wrong on this abstraction, and then the `source` of that error can tell the user what `io::Error` caused that error.

src/service.rs
```diff
@@ -524,11 +522,12 @@ impl Service {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn start<S: AsRef<OsStr>>(&self, service_arguments: &[S]) -> Result<()> {
+    pub fn start<S: AsRef<OsStr>>(&self, service_arguments: &[S]) -> Result<(), ErrorKind> {
         let wide_service_arguments = service_arguments
             .iter()
-            .map(|s| WideCString::from_str(s).chain_err(|| ErrorKind::InvalidStartArgument))
-            .collect::<Result<Vec<WideCString>>>()?;
+            .map(|s| WideCString::from_str(s).map_err(|_| ErrorKind::InvalidStartArgument))
```
faern: If you use causes as I argued in the other thread, then you don't need to discard the error here. Instead it can be included inside `InvalidStartArgument` so the user can get a nice chain of errors. Explaining more in detail what went wrong and why.

src/service.rs
```diff
@@ -12,10 +12,10 @@ use winapi::um::{winnt, winsvc};
 
 use crate::double_nul_terminated;
 use crate::sc_handle::ScHandle;
-use crate::{ErrorKind, Result, ResultExt};
+use crate::ErrorKind;
```
faern: Nitpick. But could you not just import `crate::Result` and have that handy type definition available? So you don't need to change all the return type declarations below.

src/service.rs
rustysec: definitely, i'm hacking on this in phases. thanks for the feedback.

src/service.rs
rustysec: done!

src/lib.rs
rustysec: forgot about that in the process of refactoring, thanks

src/lib.rs
```diff
@@ -172,82 +172,100 @@
...
-            display("Invalid service start type: {}", raw_value)
-        }
-        /// Invalid raw representation of [`ServiceErrorControl`].
-        InvalidServiceErrorControl(raw_value: u32) {
-            description("Invalid service error control type")
-            display("Invalid service error control type: {}", raw_value)
-        }
-    }
-    foreign_links {
-        System(::std::io::Error) #[doc = "System call error"];
```
faern: I did not check that we had lumped all `io::Error`s into one error before this PR. Sorry for picking about you just keeping the same style :/

faern:
Well, except that AppVeyor is still having errors. But after you fix those :)

src/lib.rs
rustysec: i went ahead and broke them all out, i feel like the error names i chose may be a little cumbersome so welcoming any feedback on that as well.

src/lib.rs
rustysec: no worries, i figured you just wanted the enhancement

rustysec:
fixed the tests :beers:

pronebird:
@rustysec thanks man! Good work! 🤝

PR #26, opened 2019-03-22
Break apart leaf errors

faern:
I try again to simplify/clean the error type(s) and make the error handling abstracted on a good and nice level. I'm experimenting a bit, and I'm not 100% sure I have reached the optimal design, but I'm thinking it's starting to be rather nice now.

This PR basically does two things to the errors:

1. Make all `from_raw(u32)` constructors return a unified `TryFromIntError(u32)`. Much like how the standard library implementations of `TryFrom` on integers return one single error type.

    I still need to keep two of the previous `from_raw` error variants on the combination `Error` enum. Since those can happen as part of methods with multiple types of errors in them.

    Overall I think this is cleaner. On the lower level, constructors like `ServiceStatus::from_raw` no longer return a huge error with many variants. Instead it just returns a very tiny and simple error informing that the given integer was not correct.

2. Join all errors containing an `io::Error` into one `Error::Winapi` variant. I'm aware that this goes completely against how I argued in #24 very recently. Sorry for this @rustysec. But I have an explanation. When I argued each IO error should have its own variant I was under the impression that at least one method where IO errors could happen had >1 potential source of IO errors in them. Therefore making it important to separate them, so the caller could figure out which of the multiple winapi calls failed. But upon further inspection I see now that each method does no more than one winapi call. As such, it's enough if the returned error indicate that **the** winapi call failed.

    This also makes the error abstraction level more symmetrical. An example: Now `Service::start` can error with either the `InvalidStartArgument` or the `Winapi` error variant. This is clear, either the caller gave an invalid argument, or the actual start call failed. Simple. Previously it could error with either the `InvalidStartArgument` or the `ServiceStartFailed` error variant. I feel that the `ServiceStartFailed` error should be one level higher. The `InvalidStartArgument` should be a **cause** of the `ServiceStartFailed` error, it should not be its sibling. With this restructure any caller to `Service::start` is now free to wrap the returned error in some `ServiceStartFailed` error, and the cause of failing to start it is either because the arguments were wrong, or the winapi call failed.

Please provide feedback and we'll reach something awesome together :)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/26)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -11,7 +11,7 @@ use winapi::shared::winerror::{ERROR_SERVICE_SPECIFIC_ERROR, NO_ERROR};
 use winapi::um::{winnt, winsvc};
 
 use crate::sc_handle::ScHandle;
-use crate::{double_nul_terminated, Error, Result};
+use crate::{double_nul_terminated, Error};
 
 bitflags::bitflags! {
     /// Enum describing the types of Windows services.
@@ -82,12 +82,12 @@ impl ServiceStartType {
         *self as u32
     }
 
-    pub fn from_raw(raw: u32) -> Result<ServiceStartType> {
+    pub fn from_raw(raw: u32) -> Result<ServiceStartType, TryFromIntError> {
         match raw {
             x if x == ServiceStartType::AutoStart.to_raw() => Ok(ServiceStartType::AutoStart),
             x if x == ServiceStartType::OnDemand.to_raw() => Ok(ServiceStartType::OnDemand),
             x if x == ServiceStartType::Disabled.to_raw() => Ok(ServiceStartType::Disabled),
-            _ => Err(Error::InvalidServiceStartType(raw)),
+            _ => Err(TryFromIntError(raw)),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/service_manager.rs
```diff
@@ -161,7 +159,8 @@ impl ServiceManager {
             launch_command_buffer.push(wide);
         }
 
-        let launch_command = WideCString::from_wide_str(launch_command_buffer).unwrap();
```
pronebird: I don't think it's possible for that call to fail since we double check every single buffer absorbed by `launch_command_buffer`

src/service.rs
```diff
@@ -477,16 +485,19 @@ impl ServiceStatus {
         raw_status
     }
 
-    fn from_raw(raw_status: winsvc::SERVICE_STATUS) -> Result<Self> {
+    /// Tries to parse a `SERVICE_STATUS` into a Rust [`ServiceStatus`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the `dwCurrentState` field does not represent a valid [`ServiceState`].
+    fn from_raw(raw: winsvc::SERVICE_STATUS) -> Result<Self, TryFromIntError> {
```
pronebird: I don't think `winsvc::SERVICE_STATUS` is an integer. 😄 You don't have to follow the std, you can simply Add `TryFromRawError`

src/service.rs
```diff
@@ -233,7 +233,15 @@ pub struct ServiceConfig {
...
-    pub unsafe fn from_raw(raw: winsvc::QUERY_SERVICE_CONFIGW) -> Result<ServiceConfig> {
+    /// Tries to parse a `QUERY_SERVICE_CONFIGW` into Rust [`ServiceConfig`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the `dwServiceType` field does not represent a valid
+    /// [`ServiceStartType`].
+    pub unsafe fn from_raw(
+        raw: winsvc::QUERY_SERVICE_CONFIGW,
+    ) -> Result<ServiceConfig, TryFromIntError> {
```
pronebird: Same here, `QUERY_SERVICE_CONFIGW` is actually a struct

src/service_manager.rs
faern: That is true. But the day a panic happens in production a custom message in the panic will really help finding the cause. We have not really cared a lot about this earlier, I have to admit that. But the community is fairly leaning towards always using `expect` over `unwrap`, to easier track down nasty bugs.

It might be that we one day change how we construct `launch_command_buffer` or there might even be a bug in some new version of `from_wide_str` or whatnot that might trigger a panic here.

src/service.rs
faern: Good input! I completely forgot that not all these errors are from parsing integers.. Renamed it a bit.

src/service.rs
faern: Whoops. Fixed.

PR #27, opened 2019-04-10
Document change to ServiceInfo struct

aepsil0n:
I did not find this in the changelog, but had a breaking change because of this addition in my code. Thought it makes sense to point that out in the changelog.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/27)


pronebird:
Thanks, @aepsil0n 👍

aepsil0n:
Thank you for adapting this. I actually did not realize that line meant this API change until your comment.

PR #30, opened 2019-07-24
Add support for extended control codes

hankbao:
First of all, thank you very much for this crate.

I found that it doesn't deal with extended control codes yet. So I think maybe add some support for these codes will be appropriate.

However, I'm new to rust so that any suggestion will be welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/30)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,5 +13,5 @@ edition = "2018"
 [target.'cfg(windows)'.dependencies]
 bitflags = "1.0.1"
 err-derive = "0.1.5"
-winapi = { git = "https://github.com/mullvad/winapi-rs.git", rev = "4bcf5cab87124bbeef8c1a445137494d874f8082", features = ["dbt", "std", "winbase", "winerror", "winsvc", "winuser"] }
+winapi = { git = "https://github.com/mullvad/winapi-rs.git", rev = "4bcf5cab87124bbeef8c1a445137494d874f8082", features = ["dbt", "std", "winbase", "winerror", "winsvc"] }
 widestring = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,5 +13,5 @@ edition = "2018"
 [target.'cfg(windows)'.dependencies]
 bitflags = "1.0.1"
 err-derive = "0.1.5"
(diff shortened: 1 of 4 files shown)


hankbao:
Found some pointer-related issues regard to the `Send` trait in the previous commits. Managed to fix them now.

Some limitations:

- Turns out the `SERVICE_CONTROL_DEVICEEVENT` is some bit of complicated. So I drop the support for it currently. 
- Due to some missing parts from the `winapi` crate, `SERVICE_CONTROL_TIMECHANGE` control lacks its `SERVICE_TIMECHANGE_INFO` event data
- `SERVICE_CONTROL_USERMODEREBOOT` is also not supported for the same reason

pronebird:
Hi @hankbao,

Thanks for your efforts. It's great to see more features landing in the crate. I've left a few comments on your PR.

hankbao:
@pronebird Thank you for pointing out these issues within my code. I'll address them later.

hankbao:
Hi @pronebird. I've managed to address most of the issues you pointed out.

Still, I'm not sure about how to deal with some `GUID` caused `ParseRawError`. So I leave two `FIXME` comments in the code currently.

Do you think it would be appropriate to change `ParseRawError` as an `enum` which can accept not only `u32` but also other types like `String`?

pronebird:
@hankbao Good work! I approved this PR. We can clean up then error handling in `ParseRawError` later. I'll add @faern to have the final look.

hankbao:
> @hankbao Good work! I approved this PR. We can clean up the error handling later. I'll add @faern to have the final look.

Thanks a lot. Hopefully, this PR can be merged.

hankbao:
I've updated the code according to your comments. Please continue to review. Thank you and sorry for the late reply.

hankbao:
It seems that all issues are resolved so far. Should we land this PR or is there anything I've missed?

hankbao:
@pronebird @faern This is my first rust PR, so thank you very much for your great help.

Hopefully, `SERVICE_CONTROL_DEVICEEVENT`, `SERVICE_CONTROL_TIMECHANGE` and `SERVICE_CONTROL_USERMODEREBOOT` would also be supported soon.

pronebird:
@hankbao thanks for contributing! 💪 Great work. 🎉

pronebird:
Closed via 0c60f1a8f6158cfa1fef3049e8b5aa6f95cfe2b5

PR #37, opened 2019-11-16
Treat 'ServiceAction' delay field as count of milliseconds

mvd-ows:
As above.

The output from `service_failure_actions.rs` seems to be unchanged. Which makes sense, because there was always a symmetry in how we handled this value.

I've verified the changes by looking at how the services applet in Windows interprets the delay value. With these changes, both our code and the services applet have a mutual understanding of how the value should be interpreted.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/37)


mvd-ows:
CI was upset about `Duration.as_millis()`, so pushed work-around for that.

PR #44, opened 2020-04-23
Retrieve pid when querying service status

rustysec:
Uses `QueryServiceStatusEx` to obtain service pid.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/44)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -1135,6 +1135,16 @@ impl<'a> From<&'a winsvc::SERVICE_STATUS> for ServiceExitCode {
     }
 }
 
+impl<'a> From<&'a winsvc::SERVICE_STATUS_PROCESS> for ServiceExitCode {
+    fn from(service_status: &'a winsvc::SERVICE_STATUS_PROCESS) -> Self {
+        if service_status.dwWin32ExitCode == ERROR_SERVICE_SPECIFIC_ERROR {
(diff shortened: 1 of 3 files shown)


src/service.rs
```diff
@@ -1256,6 +1270,30 @@ impl ServiceStatus {
             exit_code: ServiceExitCode::from(&raw),
             checkpoint: raw.dwCheckPoint,
             wait_hint: Duration::from_millis(raw.dwWaitHint as u64),
+            pid: None,
```
pronebird, line 1273: We tend to maintain the snakecase variants of the corresponding field names used in WinAPI, so it's easier to relate to the original WinAPI structures. The only exception is `exit_code` which is a combo of both `dwWin32ExitCode` and `dwServiceSpecificExitCode`.

So `SERVICE_STATUS_PROCESS .dwProcessId` would translate to something like `ServiceStatus.process_id`.

I am still looking at your PR but that's something that draw attention at first sight.

rustysec: Sorry I managed to miss this notification. You are completely correct on the naming consistency, I'll fix that up.

faern:
Merged! I changed the field ordering and added the feature to the changelog.

Thanks for the contribution! It's a nice improvement for sure.

PR #48, opened 2020-06-15
Implement errors without any external crate

oherrala:
It's a bit more code but now it's done :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,5 @@ default-target = "x86_64-pc-windows-msvc"
 
 [target.'cfg(windows)'.dependencies]
 bitflags = "1.2"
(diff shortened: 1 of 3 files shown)


oherrala:
> This is not equal to what we have with err-derive. `InvalidAccountName(err) => err` would maintain the same chain of errors with sources. The way it's currently implemented skips err here from the chain.

I rebased the PR on top of current master and fixed this. Is it now what it should be?

pronebird:
It's sad that this PR was left unmerged. However many things have changed in the past years and vanilla errors have finally landed in master via https://github.com/mullvad/windows-service-rs/pull/94. I assume the end result is the same and so I am going to close this PR.

Issue #45, opened 2020-05-06
Use safe MaybeUninit instead of uninitialized/zeroed


faern:
`std::mem::uninitialized` and `zeroed` were deprecated in Rust 1.39 since they were found unsound. We should upgrade this library no not use them. And instead use the new `std::mem::MaybeUninit`

faern:
Fixed on `master` since a while back. We still use `mem::zeroed` in a few places. But that has not been deprecated and should be safe. It's just uninitialized that's strongly recommended against.

PR #52, opened 2020-08-06
service: Replace std::mem::* calls by std::mem::MaybeUninit.

oktal:
I used your library in https://github.com/timberio/vector/pull/2896 to implement windows service support.

The command-line supports a `service status` sub-command. However, it looks like the `query_config` call is now crashing with a recent version of `rustc` (it was working when I initially submitted the PR).

Using `std::mem::MaybeUninit` to safely initialize raw memory seems to fix the issue.

This change should also address #45 

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -447,7 +447,6 @@ impl RawServiceInfo {
     }
 }
 
-
 /// A struct that describes the service.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct ServiceConfig {
@@ -1242,7 +1241,8 @@ pub struct ServiceStatus {
 
 impl ServiceStatus {
     pub(crate) fn to_raw(&self) -> winsvc::SERVICE_STATUS {
-        let mut raw_status = unsafe { mem::zeroed::<winsvc::SERVICE_STATUS>() };
+        let mut raw_status =
+            unsafe { mem::MaybeUninit::<winsvc::SERVICE_STATUS>::zeroed().assume_init() };
         raw_status.dwServiceType = self.service_type.bits();
         raw_status.dwCurrentState = self.current_state.to_raw();
         raw_status.dwControlsAccepted = self.controls_accepted.bits();
@@ -1273,7 +1273,6 @@ impl ServiceStatus {
             process_id: None,
         })
     }
-
 
     /// Tries to parse a `SERVICE_STATUS_PROCESS` into a Rust [`ServiceStatus`].
     ///
@@ -1393,7 +1392,8 @@ impl Service {
 
     /// Get the service status from the system.
     pub fn query_status(&self) -> crate::Result<ServiceStatus> {
-        let mut raw_status = unsafe { mem::zeroed::<winsvc::SERVICE_STATUS_PROCESS>() };
+        let mut raw_status =
+            unsafe { mem::MaybeUninit::<winsvc::SERVICE_STATUS_PROCESS>::zeroed().assume_init() };
         let mut bytes_needed: DWORD = 0;
         let success = unsafe {
             winsvc::QueryServiceStatusEx(
@@ -1424,14 +1424,17 @@ impl Service {
     /// Get the service config from the system.
     pub fn query_config(&self) -> crate::Result<ServiceConfig> {
         // As per docs, the maximum size of data buffer used by QueryServiceConfigW is 8K
-        let mut data = [0u8; MAX_QUERY_BUFFER_SIZE];
+        let mut buf = unsafe {
+            mem::MaybeUninit::<[mem::MaybeUninit<u8>; MAX_QUERY_BUFFER_SIZE]>::uninit()
+                .assume_init()
+        };
         let mut bytes_written: u32 = 0;
 
         let success = unsafe {
             winsvc::QueryServiceConfigW(
                 self.service_handle.raw_handle(),
-                data.as_mut_ptr() as _,
-                data.len() as u32,
+                buf.as_mut_ptr() as _,
+                buf.len() as u32,
                 &mut bytes_written,
             )
         };
@@ -1440,7 +1443,8 @@ impl Service {
             Err(Error::Winapi(io::Error::last_os_error()))
         } else {
             unsafe {
-                let raw_config = data.as_ptr() as *const winsvc::QUERY_SERVICE_CONFIGW;
+                let raw_config =
+                    std::mem::transmute::<_, *const winsvc::QUERY_SERVICE_CONFIGW>(buf.as_ptr());
                 ServiceConfig::from_raw(*raw_config)
             }
         }
@@ -1495,8 +1499,9 @@ impl Service {
     /// Please refer to MSDN for more info:\
     /// <https://docs.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-_service_failure_actions_flag>
     pub fn set_failure_actions_on_non_crash_failures(&self, enabled: bool) -> crate::Result<()> {
-        let mut raw_failure_actions_flag =
-            unsafe { mem::zeroed::<winsvc::SERVICE_FAILURE_ACTIONS_FLAG>() };
+        let mut raw_failure_actions_flag = unsafe {
+            mem::MaybeUninit::<winsvc::SERVICE_FAILURE_ACTIONS_FLAG>::zeroed().assume_init()
+        };
 
         raw_failure_actions_flag.fFailureActionsOnNonCrashFailures = if enabled { 1 } else { 0 };
 
@@ -1608,7 +1613,8 @@ impl Service {
     /// # }
     /// ```
     pub fn update_failure_actions(&self, update: ServiceFailureActions) -> crate::Result<()> {
-        let mut raw_failure_actions = unsafe { mem::zeroed::<winsvc::SERVICE_FAILURE_ACTIONSW>() };
+        let mut raw_failure_actions =
+            unsafe { mem::MaybeUninit::<winsvc::SERVICE_FAILURE_ACTIONSW>::zeroed().assume_init() };
 
         let mut reboot_msg = to_wide_slice(update.reboot_msg)
             .map_err(Error::InvalidServiceActionFailuresRebootMessage)?;
@@ -1640,7 +1646,8 @@ impl Service {
 
     /// Private helper to send the control commands to the system.
     fn send_control_command(&self, command: ServiceControl) -> crate::Result<ServiceStatus> {
-        let mut raw_status = unsafe { mem::zeroed::<winsvc::SERVICE_STATUS>() };
+        let mut raw_status =
+            unsafe { mem::MaybeUninit::<winsvc::SERVICE_STATUS>::zeroed().assume_init() };
         let success = unsafe {
             winsvc::ControlService(
                 self.service_handle.raw_handle(),
--- a/src/service.rs
+++ b/src/service.rs
@@ -1392,14 +1392,13 @@ impl Service {
 
     /// Get the service status from the system.
     pub fn query_status(&self) -> crate::Result<ServiceStatus> {
-        let mut raw_status =
-            unsafe { mem::MaybeUninit::<winsvc::SERVICE_STATUS_PROCESS>::zeroed().assume_init() };
+        let mut raw_status = mem::MaybeUninit::<winsvc::SERVICE_STATUS_PROCESS>::zeroed();
         let mut bytes_needed: DWORD = 0;
         let success = unsafe {
             winsvc::QueryServiceStatusEx(
                 self.service_handle.raw_handle(),
                 winsvc::SC_STATUS_PROCESS_INFO,
-                &mut raw_status as *mut _ as _,
+                raw_status.as_mut_ptr() as _,
                 std::mem::size_of::<winsvc::SERVICE_STATUS_PROCESS>() as u32,
                 &mut bytes_needed,
             )
@@ -1407,7 +1406,8 @@ impl Service {
         if success == 0 {
             Err(Error::Winapi(io::Error::last_os_error()))
         } else {
-            ServiceStatus::from_raw_ex(raw_status).map_err(Error::InvalidServiceState)
+            let status = unsafe { raw_status.assume_init() };
+            ServiceStatus::from_raw_ex(status).map_err(Error::InvalidServiceState)
         }
     }
 
@@ -1428,6 +1428,7 @@ impl Service {
             mem::MaybeUninit::<[mem::MaybeUninit<u8>; MAX_QUERY_BUFFER_SIZE]>::uninit()
                 .assume_init()
         };
+
         let mut bytes_written: u32 = 0;
 
         let success = unsafe {
@@ -1443,6 +1444,7 @@ impl Service {
             Err(Error::Winapi(io::Error::last_os_error()))
         } else {
             unsafe {
+                // Safety: buf has been initialized
                 let raw_config =
                     std::mem::transmute::<_, *const winsvc::QUERY_SERVICE_CONFIGW>(buf.as_ptr());
                 ServiceConfig::from_raw(*raw_config)
@@ -1517,10 +1519,13 @@ impl Service {
     /// Query the system for the boolean indication that the service is configured to run failure
     /// actions on non-crash failures.
     pub fn get_failure_actions_on_non_crash_failures(&self) -> crate::Result<bool> {
-        let mut data = [0u8; MAX_QUERY_BUFFER_SIZE];
+        let mut buf = unsafe {
+            mem::MaybeUninit::<[mem::MaybeUninit<u8>; MAX_QUERY_BUFFER_SIZE]>::uninit()
+                .assume_init()
+        };
 
         let raw_failure_actions_flag: winsvc::SERVICE_FAILURE_ACTIONS_FLAG = unsafe {
-            self.query_config2(winsvc::SERVICE_CONFIG_FAILURE_ACTIONS_FLAG, &mut data)
+            self.query_config2(winsvc::SERVICE_CONFIG_FAILURE_ACTIONS_FLAG, &mut buf)
                 .map_err(Error::Winapi)?
         };
 
@@ -1552,10 +1557,12 @@ impl Service {
     /// Query the configured failure actions for the service.
     pub fn get_failure_actions(&self) -> crate::Result<ServiceFailureActions> {
         unsafe {
-            let mut data = [0u8; MAX_QUERY_BUFFER_SIZE];
+            let mut buf =
+                mem::MaybeUninit::<[mem::MaybeUninit<u8>; MAX_QUERY_BUFFER_SIZE]>::uninit()
+                    .assume_init();
 
             let raw_failure_actions: winsvc::SERVICE_FAILURE_ACTIONSW = self
-                .query_config2(winsvc::SERVICE_CONFIG_FAILURE_ACTIONS, &mut data)
+                .query_config2(winsvc::SERVICE_CONFIG_FAILURE_ACTIONS, &mut buf)
                 .map_err(Error::Winapi)?;
 
             ServiceFailureActions::from_raw(raw_failure_actions)
@@ -1646,20 +1653,21 @@ impl Service {
 
     /// Private helper to send the control commands to the system.
     fn send_control_command(&self, command: ServiceControl) -> crate::Result<ServiceStatus> {
-        let mut raw_status =
-            unsafe { mem::MaybeUninit::<winsvc::SERVICE_STATUS>::zeroed().assume_init() };
+        let mut raw_status = mem::MaybeUninit::<winsvc::SERVICE_STATUS>::zeroed();
         let success = unsafe {
             winsvc::ControlService(
                 self.service_handle.raw_handle(),
                 command.raw_service_control_type(),
-                &mut raw_status,
+                raw_status.as_mut_ptr(),
             )
         };
 
         if success == 0 {
             Err(Error::Winapi(io::Error::last_os_error()))
         } else {
-            ServiceStatus::from_raw(raw_status).map_err(Error::InvalidServiceState)
+            // Safety: raw_status has been initialized
+            let status = unsafe { raw_status.assume_init() };
+            ServiceStatus::from_raw(status).map_err(Error::InvalidServiceState)
         }
     }
 
@@ -1667,7 +1675,7 @@ impl Service {
     unsafe fn query_config2<T: Copy>(
         &self,
         kind: DWORD,
-        data: &mut [u8; MAX_QUERY_BUFFER_SIZE],
+        data: &mut [mem::MaybeUninit<u8>],
     ) -> io::Result<T> {
         let mut bytes_written: u32 = 0;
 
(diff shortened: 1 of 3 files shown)


Hoverbear:
@oktal Do you think you could help fixup the CI here? Then we can bug @faern to take a look. :)

src/service.rs
```diff
@@ -1393,7 +1392,8 @@ impl Service {
 
     /// Get the service status from the system.
     pub fn query_status(&self) -> crate::Result<ServiceStatus> {
-        let mut raw_status = unsafe { mem::zeroed::<winsvc::SERVICE_STATUS_PROCESS>() };
+        let mut raw_status =
+            unsafe { mem::MaybeUninit::<winsvc::SERVICE_STATUS_PROCESS>::zeroed().assume_init() };
```
faern, line 1396: The docs for `mem::zeroed()` states that it has the same effect as what you changed this into. So this makes no difference in safety or correctness.

The way we probably want to use `MaybeUninit` is with `MaybeUninit::uninit()` and then getting a pointer to it with `raw_status.as_mut_ptr()` in order to pass it over FFI to be filled in. And then after it has been filled in call `assume_init()` on it. Like how the docs describe it in https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html. The point is to keep the actual C struct inside the `MaybeUninit` until it has been properly initialized, and then take it out with `assume_init()`. If we take it out before it's initialized `MaybeUninit` does not help us much.

src/service.rs
```diff
@@ -1424,14 +1424,17 @@ impl Service {
     /// Get the service config from the system.
     pub fn query_config(&self) -> crate::Result<ServiceConfig> {
         // As per docs, the maximum size of data buffer used by QueryServiceConfigW is 8K
-        let mut data = [0u8; MAX_QUERY_BUFFER_SIZE];
+        let mut buf = unsafe {
+            mem::MaybeUninit::<[mem::MaybeUninit<u8>; MAX_QUERY_BUFFER_SIZE]>::uninit()
+                .assume_init()
```
faern, line 1429: This was safe from the start. `[u8; N]` is a safely initialized array. Nothing around here has potential uninitialized UB.

src/service.rs
oktal: That makes sense, I'll update the PR accordingly, then

src/service.rs
oktal: Then I'm not sure why I'm getting a segmentation fault when attempting to run the `service_config` example

```rust
 /// Get the service config from the system.
    pub fn query_config(&self) -> crate::Result<ServiceConfig> {
        // As per docs, the maximum size of data buffer used by QueryServiceConfigW is 8K
        let mut buf = [0u8; MAX_QUERY_BUFFER_SIZE];
        let mut bytes_written: u32 = 0;

        let success = unsafe {
            winsvc::QueryServiceConfigW(
                self.service_handle.raw_handle(),
                buf.as_mut_ptr() as _,
                MAX_QUERY_BUFFER_SIZE as u32,
                &mut bytes_written,
            )
        };

        if success == 0 {
            Err(Error::Winapi(io::Error::last_os_error()))
        } else {
            unsafe {
                // Safety: buf has been initialized
                let raw_config =
                    std::mem::transmute::<_, *const winsvc::QUERY_SERVICE_CONFIGW>(buf.as_ptr());
                ServiceConfig::from_raw(*raw_config)
            }
        }
    }
```

```
cargo run --example service_config
   Compiling windows-service v0.3.0 (C:\Dev\windows-service-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 1.68s
     Running `target\debug\examples\service_config.exe`
error: process didn't exit successfully: `target\debug\examples\service_config.exe` (exit code: 1)
```

Using `MaybeUninit` fixes the issue

```rust
/// Get the service config from the system.
    pub fn query_config(&self) -> crate::Result<ServiceConfig> {
        // As per docs, the maximum size of data buffer used by QueryServiceConfigW is 8K
        let mut buf = mem::MaybeUninit::<[u8; MAX_QUERY_BUFFER_SIZE]>::uninit();
        let mut bytes_written: u32 = 0;

        let success = unsafe {
            winsvc::QueryServiceConfigW(
                self.service_handle.raw_handle(),
                buf.as_mut_ptr() as _,
                MAX_QUERY_BUFFER_SIZE as u32,
                &mut bytes_written,
            )
        };

        if success == 0 {
            Err(Error::Winapi(io::Error::last_os_error()))
        } else {
            unsafe {
                // Safety: buf has been initialized
                let raw_config =
                    std::mem::transmute::<_, *const winsvc::QUERY_SERVICE_CONFIGW>(buf.as_ptr());
                ServiceConfig::from_raw(*raw_config)
            }
        }
    }
```

```
cargo run --example service_config
   Compiling windows-service v0.3.0 (C:\Dev\windows-service-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 0.56s
     Running `target\debug\examples\service_config.exe`
ServiceConfig {
    service_type: SHARE_PROCESS,
    start_type: AutoStart,
    error_control: Normal,
    executable_path: "C:\\WINDOWS\\system32\\lsass.exe",
    load_order_group: Some(
        "MS_WindowsRemoteValidation",
    ),
    tag_id: 0,
    dependencies: [
        Service(
            "LanmanWorkstation",
        ),
    ],
    account_name: Some(
        "LocalSystem",
    ),
    display_name: "Netlogon",
}
```

Any idea ?

pronebird: It looks like when using the `QueryServiceConfigW` API we should be able to get away with initializing `QUERY_SERVICE_CONFIGW` directly as opposed to using the 8k byte array. I think the docs for `QueryServiceConfigW` are copy-and-paste from the docs for `QueryServiceConfig2W` which accepts a flag that controls the kind of information that the generic `lpBuffer` parameter is being filled with. In case of `QueryServiceConfigW` we have a strictly typed buffer.

So either way I think it's reasonable to use `MaybeUninit` here instated of a large byte buffer:

```rs
let mut buf = unsafe {  mem::MaybeUninit::<winsvc::QUERY_SERVICE_CONFIGW>::zeroed() };
```

Obviously the size of the `QUERY_SERVICE_CONFIGW` should be used in place of `MAX_QUERY_BUFFER_SIZE` too.

Docs:

1. https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-queryserviceconfigw
2. https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-queryserviceconfig2w

oktal: Looking at the example from the documentation, I don't think that's the case.

The MSDN example is first retrieving the number of bytes needed prior to calling `QueryServiceConfig` again:

```cpp
   if( !QueryServiceConfig( 
        schService, 
        NULL, 
        0, 
        &dwBytesNeeded))
    {
        dwError = GetLastError();
        if( ERROR_INSUFFICIENT_BUFFER == dwError )
        {
            cbBufSize = dwBytesNeeded;
            lpsc = (LPQUERY_SERVICE_CONFIG) LocalAlloc(LMEM_FIXED, cbBufSize);
        }
        else
        {
            printf("QueryServiceConfig failed (%d)", dwError);
            goto cleanup; 
        }
    }
```

It looks like the number of bytes needed is not `mem::size_of<QUERY_SERVICE_CONFIGW>()` but really depends on the configuration:

```console
λ cargo run --example service_config -- netlogon                
   Compiling windows-service v0.3.0 (C:\Dev\windows-service-rs) 
    Finished dev [unoptimized + debuginfo] target(s) in 1.39s   
     Running `target\debug\examples\service_config.exe netlogon`
bytes_needed = 354                                              
ServiceConfig {                                                 
    service_type: SHARE_PROCESS,                                
    start_type: AutoStart,                                      
    error_control: Normal,                                      
    executable_path: "C:\\WINDOWS\\system32\\lsass.exe",        
    load_order_group: Some(                                     
        "MS_WindowsRemoteValidation",                           
    ),                                                          
    tag_id: 0,                                                  
    dependencies: [                                             
        Service(                                                
            "LanmanWorkstation",                                
        ),                                                      
    ],                                                          
    account_name: Some(                                         
        "LocalSystem",                                          
    ),                                                          
    display_name: "Netlogon",                                   
}                                                               
```

```console
λ cargo run --example service_config -- SysMain                                               
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s                                 
     Running `target\debug\examples\service_config.exe SysMain`                               
bytes_needed = 394                                                                            
ServiceConfig {                                                                               
    service_type: SHARE_PROCESS,                                                              
    start_type: AutoStart,                                                                    
    error_control: Ignore,                                                                    
    executable_path: "C:\\WINDOWS\\system32\\svchost.exe -k LocalSystemNetworkRestricted -p", 
    load_order_group: Some(                                                                   
        "profsvc_group",                                                                      
    ),                                                                                        
    tag_id: 0,                                                                                
    dependencies: [                                                                           
        Service(                                                                              
            "rpcss",                                                                          
        ),                                                                                    
        Service(                                                                              
            "fileinfo",                                                                       
        ),                                                                                    
    ],                                                                                        
    account_name: Some(                                                                       
        "LocalSystem",                                                                        
    ),                                                                                        
    display_name: "SysMain",                                                                  
}                                                                                             
```

So I think I would keep the 8k byte array approach.

pronebird: Alright. That's possible. In the end they might be returning something that looks like `QUERY_SERVICE_CONFIGW` but is actually a different possibly private struct that just happens to share the same fields.

oktal:
Hi @faern @pronebird, PR has been updated.

I tried to use `MaybeUninit` in a more idiomatic way, following the docs. Let me know if you have any issue

pronebird (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @faern)

faern (changes requested):
> Reviewed 1 of 2 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @oktal and @pronebird)

___
*[src/service.rs, line 1396 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-ME2cf3D8R6RWmPg81O0-r1-1396:-MJaqYBm6OhcsiSRCSNh:b-sgpyv3) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5db3ff216d70f676c85eb8acf2e912d6da0c8fb0/src/service.rs#L1396)):*
<details><summary><i>Previously, oktal (Mathieu Stefani) wrote…</i></summary><blockquote>

That makes sense, I'll update the PR accordingly, then
</blockquote></details>

This looks more correct! :+1:
___
*[src/service.rs, line 1429 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-ME2cf3D8R6RWmPg81O0-r1-1429:-MJat2y2-8WOTcnVkuZB:b-5a2t7m) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5db3ff216d70f676c85eb8acf2e912d6da0c8fb0/src/service.rs#L1429)):*
The struct `QUERY_SERVICE_CONFIGW` contains pointers. I think the data those pointers point to are written to the buffer *after* the initial `QUERY_SERVICE_CONFIGW` content. That's why you send in a buffer and not just a pointer to a `QUERY_SERVICE_CONFIGW`. It uses that buffer to store the `QUERY_SERVICE_CONFIGW` (in the beginning of the buffer) and the strings it points to just after said buffer. So it becomes kind of "self referential" in a way. That way the caller can just free the buffer and automatically free both the struct and the .

It should be possible to verify the above by printing out the buffer pointer and the actual pointers returned and doing some light math on that to see if they point within the 8k region the sent in pointer points to.

> Obviously the size of the QUERY_SERVICE_CONFIGW should be used in place of MAX_QUERY_BUFFER_SIZE too.

If what I wrote above is true, that's why you need to send in `MAX_QUERY_BUFFER_SIZE` and not `size_of::<QUERY_SERVICE_CONFIGW>()`.

> Then I'm not sure why I'm getting a segmentation fault when attempting to run the service_config example

Do you really get a segfault? All I see is `error: process didn't exit successfully: target\debug\examples\service_config.exe (exit code: 1)` which does not look like a segfault. It looks like a normal `Err(_)` of some kind but we can't see what the actual error was. Try writing out the `Display` impl for your error and all it's causes/children.
___
*[src/service.rs, line 1449 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-MJaro0PA-nwAQf58zFV:-MJaro0PA-nwAQf58zFW:b92z6l9) ([raw file](https://github.com/mullvad/windows-service-rs/blob/af362c48a869358d697dde062c9dd08eaaaa44d6/src/service.rs#L1449)):*
> ```Rust
>                 // Safety: buf has been initialized
>                 let raw_config =
>                     std::mem::transmute::<_, *const winsvc::QUERY_SERVICE_CONFIGW>(buf.as_ptr());
> ```

`transmute` is not needed ta cast between raw pointer types. I don't understand why this part of the change is needed.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @faern and @oktal)

___
*[src/service.rs, line 1429 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-ME2cf3D8R6RWmPg81O0-r1-1429:-MJkkSfL2tUi0qRdzHpT:b9oyyyy) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5db3ff216d70f676c85eb8acf2e912d6da0c8fb0/src/service.rs#L1429)):*

The struct `QUERY_SERVICE_CONFIGW` contains pointers. I think the data those pointers point to are written to the buffer *after* the initial `QUERY_SERVICE_CONFIGW` content. That's why you send in a buffer and not just a pointer to a `QUERY_SERVICE_CONFIGW`. It uses that buffer to store the `QUERY_SERVICE_CONFIGW` (in the beginning of the buffer) and the strings it points to just after said buffer. So it becomes kind of "self referential" in a way. That way the caller can just free the buffer and automatically free both the struct and the .

It should be possible to verify the above by printing out the buffer pointer and the actual pointers returned and doing some light math on that to see if they point within the 8k region the sent in pointer points to.

> Obviously the size of the QUERY_SERVICE_CONFIGW should be used in place of MAX_QUERY_BUFFER_SIZE too.

If what I wrote above is true, that's why you need to send in `MAX_QUERY_BUFFER_SIZE` and not `size_of::<QUERY_SERVICE_CONFIGW>()`.

> Then I'm not sure why I'm getting a segmentation fault when attempting to run the service_config example

Do you really get a segfault? All I see is `error: process didn't exit successfully: target\debug\examples\service_config.exe (exit code: 1)` which does not look like a segfault. It looks like a normal `Err(_)` of some kind but we can't see what the actual error was. Try writing out the `Display` impl for your error and all it's causes/children.
</blockquote></details>

I think @faern is correct in his assumption. The buffer that the WINAPI returns contains the following bytes:

1. The`QUERY_SERVICE_CONFIGW` struct with all pointer fields pointing to the within the returned buffer.
1. The data referenced inside `QUERY_SERVICE_CONFIGW` struct.

Hence there are two ways to approach this: 

_Long path_

1. Ask WINAPI how much space it needs for `QUERY_SERVICE_CONFIGW`
1. Allocate the buffer and call back providing the allocated buffer
1. Repeat the whole sequence on `ERROR_INSUFFICIENT_BUFFER` until the sequence succeeds.

_Short path_

Pre-allocate the buffer of the max possible size `MAX_QUERY_BUFFER_SIZE` since WINAPI guarantees that the returned buffer cannot be longer than this.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @faern and @oktal)

___
*[src/service.rs, line 1663 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-MJknv-L1xqwxrYBmhJc:-MJknv-L1xqwxrYBmhJd:b-n7jdi) ([raw file](https://github.com/mullvad/windows-service-rs/blob/309f842f4a8e25c4fd28d1b006788e2f4e93ddcb/src/service.rs#L1663)):*
> ```Rust
>         &self,
>         kind: DWORD,
>         data: &mut [u8; MAX_QUERY_BUFFER_SIZE],
> ```

I haven't touched any of this in a while. But what's wrong with passing a fixed array? What does it matter what it contains given that we write into it.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @faern and @oktal)

___
*[src/service.rs, line 1515 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-MJkoM7zCCCP6rs0J0g4:-MJkoM7zCCCP6rs0J0g5:bady4hh) ([raw file](https://github.com/mullvad/windows-service-rs/blob/309f842f4a8e25c4fd28d1b006788e2f4e93ddcb/src/service.rs#L1515)):*
> ```Rust
>     /// actions on non-crash failures.
>     pub fn get_failure_actions_on_non_crash_failures(&self) -> crate::Result<bool> {
>         let mut data = [0u8; MAX_QUERY_BUFFER_SIZE];
> ```

Same here. It's perfectly normal to have a fixed size array filled with zeroes.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: 2 of 3 files reviewed, 5 unresolved discussions (waiting on @faern and @oktal)

___
*[src/service.rs, line 1550 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-MJkogNh3TlxA2I2EebZ:-MJkogNh3TlxA2I2Eeb_:b-k2pczf) ([raw file](https://github.com/mullvad/windows-service-rs/blob/309f842f4a8e25c4fd28d1b006788e2f4e93ddcb/src/service.rs#L1550)):*
> ```Rust
>     pub fn get_failure_actions(&self) -> crate::Result<ServiceFailureActions> {
>         unsafe {
>             let mut data = [0u8; MAX_QUERY_BUFFER_SIZE];
> ```

Here too.

pronebird (commented):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: all files reviewed, 5 unresolved discussions (waiting on @faern and @oktal)

___
*[examples/service_config.rs, line 20 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-MJkpPVbAUC-DkNRzLTp:-MJkpPVbAUC-DkNRzLTq:btu40jg) ([raw file](https://github.com/mullvad/windows-service-rs/blob/af362c48a869358d697dde062c9dd08eaaaa44d6/examples/service_config.rs#L20)):*
> ```Rust
> 
>     let config = service.query_config()?;
>     let status = service.query_status()?;
> ```

We do feature the use of `query_status` in `uninstall_service.rs`. We sure could have it here, but it feels as a superficial addition. Maybe not?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: all files reviewed, 5 unresolved discussions (waiting on @oktal and @pronebird)

___
*[src/service.rs, line 1429 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-ME2cf3D8R6RWmPg81O0-r1-1429:-MKJeGMw8Fnwa2MmZ6WE:b-7sfyav) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5db3ff216d70f676c85eb8acf2e912d6da0c8fb0/src/service.rs#L1429)):*

I think @faern is correct in his assumption. The buffer that the WINAPI returns contains the following bytes:

1. The`QUERY_SERVICE_CONFIGW` struct with all pointer fields pointing to the within the returned buffer.
1. The data referenced inside `QUERY_SERVICE_CONFIGW` struct.

Hence there are two ways to approach this: 

_Long path_

1. Ask WINAPI how much space it needs for `QUERY_SERVICE_CONFIGW`
1. Allocate the buffer and call back providing the allocated buffer
1. Repeat the whole sequence on `ERROR_INSUFFICIENT_BUFFER` until the sequence succeeds.

_Short path_

Pre-allocate the buffer of the max possible size `MAX_QUERY_BUFFER_SIZE` since WINAPI guarantees that the returned buffer cannot be longer than this.
</blockquote></details>

I definitely prefer the short path. Allocating 8k is not a lot in any measurable way. The cost of two subsequest syscalls must be way higher. Not to mention the added code complexity. 
___
*[src/service.rs, line 1515 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-MJkoM7zCCCP6rs0J0g4:-MKJeWE-5up6j7ZnpqoL:b-mx6043) ([raw file](https://github.com/mullvad/windows-service-rs/blob/309f842f4a8e25c4fd28d1b006788e2f4e93ddcb/src/service.rs#L1515)):*

Same here. It's perfectly normal to have a fixed size array filled with zeroes.
</blockquote></details>

Exactly. These buffers were initialized safely (otherwise there would be a bug in Rust, because this code clearly initialized it without any `unsafe {}`). `mem::uninitialized::<u8>::()` actually IS undefined behaviour. No, there are no possible byte values that don't represent a valid integer. But Rust has not defined it yet, so for now it's still undefined behavior. However, that's not a problem here, because we are not creating it uninitialized, we are initializing it with zeros.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/52)* status: all files reviewed, 5 unresolved discussions (waiting on @faern, @oktal, and @pronebird)

___
*[src/service.rs, line 1429 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/52#-ME2cf3D8R6RWmPg81O0-r1-1429:-MKLhSaxELPpHLOnBAm0:b-dl67gq) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5db3ff216d70f676c85eb8acf2e912d6da0c8fb0/src/service.rs#L1429)):*

I definitely prefer the short path. Allocating 8k is not a lot in any measurable way. The cost of two subsequest syscalls must be way higher. Not to mention the added code complexity. 
</blockquote></details>

I confirm that I witness the seg fault when using current master. My assumption that the problem is that we use the stack allocated memory vs. heap allocated memory when calling to `QueryServiceConfig*` family of WINAPI functions.

I have submitted a PR that fixes this issue:
https://github.com/mullvad/windows-service-rs/pull/55

faern:
@oktal I don't think this PR is needed any more. We did not fully agree on some of the stuff if it's needed or not. Then @pronebird did some research and got #55 merged that I think fixed all the issues you were seeing. Should we close this or is there something here that you still think we could use in `master` before publishing a release?

faern:
Closing this since the bug is fixed and no response from the author.

PR #58, opened 2020-11-23
Add `ServiceManagerAccess::ALL_ACCESS`

pravic:
for service stop or delete operations.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/58)

### DIFF
--- a/src/service_manager.rs
+++ b/src/service_manager.rs
@@ -19,6 +19,9 @@ bitflags::bitflags! {
 
         /// Can enumerate services or receive notifications.
         const ENUMERATE_SERVICE = winsvc::SC_MANAGER_ENUMERATE_SERVICE;
+
+        /// All access. Used for stopping or deleting services.
+        const ALL_ACCESS = winsvc::SC_MANAGER_ALL_ACCESS;
     }
 }
 


src/service_manager.rs
```diff
@@ -19,6 +19,9 @@ bitflags::bitflags! {
 
         /// Can enumerate services or receive notifications.
         const ENUMERATE_SERVICE = winsvc::SC_MANAGER_ENUMERATE_SERVICE;
+
+        /// All access. Used for stopping or deleting services.
+        const ALL_ACCESS = winsvc::SC_MANAGER_ALL_ACCESS;
```
faern, line 24: Do you have a concrete example what kind of code would need this access to work? Stopping and deleting services can already be done via `Service::stop` and `Service::delete` and the access is given by `ServiceAccess::STOP | ServiceAccess::DELETE`. See https://github.com/mullvad/windows-service-rs/blob/78ca40eba233b3f54bc89566233aa33b2f05a296/examples/uninstall_service.rs.

pravic:
@faern `SC_MANAGER_ALL_ACCESS` is taken from https://docs.microsoft.com/en-us/windows/win32/services/stopping-a-service

faern:
Better documentation would definitely not hurt this crate! The best would be if users can use this crate without resorting to Microsoft's C++ documentation of course. So if `ALL_ACCESS` is actually not needed, but it's unclear how to stop a service we'd gladly accept improved documentation.

I'm not sure I understand how `ALL_ACCESS` is an alternative to better docs? How does it become easier to understand how to stop a service just by adding a permission that's not actually needed to stop said service? :thinking: 

I'm not against adding `ALL_ACCESS` if it has any practical benefit. I don't know all the possible use cases. If there are things currently not possible to do with this library that adding that permission would allow doing, we'll of course add it. But I'm not sure about how the documentation is currently worded, since it seems what it talks about should already be possible.

src/service_manager.rs
pronebird: If we look at the [source code](https://github.com/tpn/winsdk-7/blob/master/v7.1A/Include/WinSvc.h#L145-L151) (win7 sdk but probably not much has changed since then), the only difference is the presence of `STANDARD_RIGHTS_REQUIRED`:

```
#define SC_MANAGER_ALL_ACCESS          (STANDARD_RIGHTS_REQUIRED      | \
                                        SC_MANAGER_CONNECT            | \
                                        SC_MANAGER_CREATE_SERVICE     | \
                                        SC_MANAGER_ENUMERATE_SERVICE  | \
                                        SC_MANAGER_LOCK               | \
                                        SC_MANAGER_QUERY_LOCK_STATUS  | \
                                        SC_MANAGER_MODIFY_BOOT_CONFIG)
```

I am not really sure what `STANDARD_RIGHTS_REQUIRED` gives. I also think that it's not necessary to request all of the permissions from SCM when you only need a few of them. However since we use bitflags-rs, you could perhaps just use `ServiceManagerAccess::all()` instead? See https://docs.rs/bitflags/1.2.1/bitflags/#methods-1

pravic: > Do you have a concrete example what kind of code would need this access to work?

@faern Besides https://docs.microsoft.com/en-us/windows/win32/services/stopping-a-service - I don't. Moreover, I've tried stopping a service with `CONNECT` and it worked fine. So, Now I am not sure.

pravic: > I think the confusion comes from the fact that there are two separate access masks

@pronebird No, that's no the case - see the MSDN example regarding stopping a service.

faern: From what I can see in the linked MSDN example it does not explicitly say that you need `ALL_ACCESS` they just happen to use it in the code. And you have already made it work for yourself with `CONNECT`. So I don't think we need this?

pravic: I'd argue that we do need this. At least, for the sake of consistency.

faern: Then at least give it better documentation. We have both proven to ourselves that it's not needed for stopping or deleting services. So there must be more fitting documentation for it. Can you find some good source on what it's intended for? I'm not very keen on adding a constant we don't know who might need for what and when.

pronebird: @pravic 

> No, that's no the case - see the MSDN example regarding stopping a service.

I saw it. I don't see anything we don't have. See the example which is a part of the repo. It demonstrates how to stop and delete the service.

pravic:
Okay

PR #59, opened 2020-11-23
Do not escape binary path for a kernel driver.

### LINKED ISSUE #57: Kernel driver path must not be quoted
Even if it contains spaces.

Valid paths:

```
%SystemRoot%\System32\drivers\processr.sys
\??\C:\Windows\gdrv2.sys
\SystemRoot\System32\drivers\1394ohci.sys
\??\C:\ProgramData\Microsoft\Windows Defender\Definition Updates\{D2115BAD-24C0-4CA0-A754-F66B01950E8A}\MpKslDrv.sys
```

Invalid (if we use [`escape_wide()`](https://github.com/mullvad/windows-service-rs/blob/78ca40eba233b3f54bc89566233aa33b2f05a296/src/service.rs#L1762)):

```
"C:\ProgramData\Microsoft\Windows Defender\Definition Updates\{D2115BAD-24C0-4CA0-A754-F66B01950E8A}\MpKslDrv.sys"
```

Source: https://stackoverflow.com/questions/50954450/createservice-and-quotes-for-lpbinarypathname-parameter#comment88914215_50954450 (could not find a better one).

pravic:
fixes #57.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/59)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -205,7 +205,7 @@ pub enum Error {
     #[error(display = "Invalid launch argument at index {}", _0)]
     InvalidLaunchArgument(usize, #[error(source)] widestring::NulError<u16>),
 
-    /// Invalid launch arguments.
+    /// Launch arguments are not supported for kernel drivers.
     #[error(display = "Kernel drivers do not support launch arguments")]
     LaunchArgumentsNotSupported,
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -205,6 +205,10 @@ pub enum Error {
     #[error(display = "Invalid launch argument at index {}", _0)]
     InvalidLaunchArgument(usize, #[error(source)] widestring::NulError<u16>),
 
+    /// Invalid launch arguments.
+    #[error(display = "Kernel drivers do not support launch arguments")]
+    LaunchArgumentsNotSupported,
+
     /// Invalid dependency name.
     #[error(display = "Invalid dependency name")]
     InvalidDependency(#[error(source)] widestring::NulError<u16>),
--- a/src/service.rs
+++ b/src/service.rs
@@ -411,6 +411,10 @@ impl RawServiceInfo {
         let mut launch_command_buffer = WideString::new();
         if service_info.service_type.intersects(ServiceType::KERNEL_DRIVER | ServiceType::FILE_SYSTEM_DRIVER) {
             // drivers do not support launch arguments
+            if !service_info.launch_arguments.is_empty() {
+                return Err(Error::LaunchArgumentsNotSupported);
+            }
+
             // also the path must not be quoted even if it contains spaces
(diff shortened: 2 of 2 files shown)


src/service.rs
```diff
@@ -407,19 +407,25 @@ impl RawServiceInfo {
...
-            escape_wide(&service_info.executable_path).map_err(Error::InvalidExecutablePath)?;
-
+        // escape executable path and arguments and combine them into a single command
         let mut launch_command_buffer = WideString::new();
-        launch_command_buffer.push(executable_path);
+        if service_info.service_type.intersects(ServiceType::KERNEL_DRIVER | ServiceType::FILE_SYSTEM_DRIVER) {
+            // drivers do not support launch arguments
+            // also the path must not be quoted even if it contains spaces
+            let executable_path = WideCString::from_os_str(&service_info.executable_path).map_err(Error::InvalidExecutablePath)?;
+            launch_command_buffer.push(executable_path.to_ustring());
```
faern, line 416: Since arguments are not supported, I think we should probably return an error in this branch if `!service_info.launch_arguments.is_empty()`. Right? Because passing arguments is an invalidly formed `ServiceInfo` and silently just discarding them can lead to hard to debug situations for users of this crate. This would justify a new separate error variant.

faern (changes requested):

src/service.rs
pravic: Changed, see a6e28b2.

faern: Great! Looks good to me.

faern (approved):
Thanks for the contribution. We have not used this for drivers ourselves. But the SO answers you linked support your case at least. If you have tested this and it works for you then I'm fine with merging it.

src/lib.rs
```diff
@@ -205,6 +205,10 @@ pub enum Error {
     #[error(display = "Invalid launch argument at index {}", _0)]
     InvalidLaunchArgument(usize, #[error(source)] widestring::NulError<u16>),
 
+    /// Invalid launch arguments.
```
pronebird, line 208: Maybe change the description?

pravic: To what? :)

pronebird: For example to "Launch arguments are not supported"?

faern: Just use the same string as the display implementation. It's even more clear.

pravic: Done in e592829.

faern:
Merged. Thanks for the contribution! :tada:

pravic:
@faern Thanks! Will a new crate version be released as well?

faern:
Do you need this right away? The `master` branch is currently breaking due to some changes into `impl Trait`. So we would need to do a breaking version bump. If there are more breaking things in the pipe it would be awesome to get them done first. I'll see what we can do.

Issue #64, opened 2021-07-12
Sync for ServiceStatusHandle


td202:
`ServiceStatusHandle` is already `Send`, but it seems it could also be `Sync`, since the [`SetServiceStatus` API is thread-safe](https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-setservicestatus). This could be useful, for instance, to allow the service control handler to set a pending state, asynchronously signal the main thread, and return.

pronebird:
Agree. Right now we allow it to be cloned, but I see no reason why we shouldn't share it across threads. Please feel free to send a PR.

https://github.com/mullvad/windows-service-rs/blob/master/src/service_control_handler.rs#L35

PR #65, opened 2021-07-15
Make `ServiceStatusHandle` Sync.

td202:
Mark `ServiceStatusHandle` as Sync, per https://github.com/mullvad/windows-service-rs/issues/64.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/65)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   `ServiceManager::local_computer()`, `ServiceManager::remote_computer()` now take 
   `impl AsRef<OsStr>` arguments.
 - Upgrade err-derive dependency to 0.3.0
+- `ServiceStatusHandle` is now Sync.
 
 ### Fixed
 - Don't escape binary path for kernel drivers as they don't support that.
--- a/src/service_control_handler.rs
+++ b/src/service_control_handler.rs
@@ -33,6 +33,7 @@ impl ServiceStatusHandle {
 // See remarks section for more info:
 // https://msdn.microsoft.com/en-us/library/windows/desktop/ms686241(v=vs.85).aspx
 unsafe impl Send for ServiceStatusHandle {}
+unsafe impl Sync for ServiceStatusHandle {}
 
 /// Abstraction over the return value of service control handler.
 /// The meaning of each of variants in this enum depends on the type of received event.


faern:
`ServiceStatusHandle` implements `Copy` so does it really need to be `Sync`? Any time you use it by value it will be automatically copied to a new instance. So you can send any number of owned instances between threads instead of just references.

Can you show a use case of something you can't do today that you could with `Sync`?

pronebird:
@faern as discussed in https://github.com/mullvad/windows-service-rs/issues/64 it looks like `ServiceStatusHandle` is thread safe, so it can be shared across threads without the need to copy.

faern:
The copying happens automatically. It's pretty uncommon to even take references to `Copy` types since the compiler can just automatically `memcpy` it instead. There is even a Clippy lint *against* calling stuff by reference on small enough copy types. Please mind that `Clone` is something different.

If you share stuff across threads they have to be `'static` most of the time anyway. And lifetimes always complicate the code a lot. So why would you want to share a reference over getting a copied instance that does not imply any lifetime constraints? Any code taking a copy will be a lot cleaner than trying to do the same with a reference.

td202:
The use case I have in mind is when the service event handler requires access to the `ServiceStatusHandle`. Because the handler is passed as an argument to `register`, which in turn creates the `ServiceStatusHandle`, the handle cannot simply be passed by value to the handler. The handle could be passed, for instance, by sharing an `AtomicRef` with the handler that will hold the `ServiceStatusHandle`.

faern:
I guess you mean that you want to have an `&AtomicRef<'a, ServiceStatusHandle>` in your `event_handler` closure and access it? Yeah OK that type won't be `Sync` indeed and can't be shared with the closure. I guess that's a valid use case! But on the other hand, if you want access to the `ServiceStatusHandle` I guess you *need* it in all closure invocations? It would be unfortunate if you need to unwrap the atomic optional since you can't guarantee it arrived yet. Then you have to handle the case of it not existing in the closure. I would maybe have used some kind of message passing channel to send the handle into the closure. And the closure blocks on the channel to wait for the first (and only) message.

We can indeed merge this PR. I just wanted to know there was a real use case.

faern:
I have published version `0.4.0` with this change.

td202:
If the `AtomicRef` is set before `set_service_status` is ever called, then it should be safe to assume that it is set in the event handler, with the (possible) exception of it being called with the `Interrogate` control event. (To be safe, I would likely return `NotImplemented` rather than panic if the `AtomicRef` is not set.)

pronebird:
@td202 I think it would be wise to guard from such edge cases and simply block the event handler thread until the ref has arrived, which should really happen fast enough, i.e:

```
let shared_ref: AtomicRef<..>
let status_handle = register(move || ... )
// set shared_ref
```

I think we should provide some abstraction for that and funnel the status handle back into the closure passed to `register`. On the other side, it's worth to note that at least in our other project (`mullvad-daemon`) we have a wrapper around `ServiceStatusHandle` that increments the counter and does all other things helping us to report the service status as it transitions through its lifecycle.

PR #66, opened 2021-11-02
chore: update widestring dependency to 0.5

vthib:
I've updated widestring to 0.5, it builds correctly and passes the tests.
It would be great if a new release with this update could be published, thanks in advance!

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/66)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 default-target = "x86_64-pc-windows-msvc"
 
 [target.'cfg(windows)'.dependencies]
-bitflags = "1.2"
+bitflags = "1.3"
 err-derive = "0.3.0"
 winapi = { version = "0.3.9", features = ["dbt", "std", "winbase", "winerror", "winsvc"] }
-widestring = "0.4.0"
+widestring = "0.5.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ keywords = ["windows", "service", "daemon"]
 categories = ["api-bindings", "os::windows-apis"]
 repository = "https://github.com/mullvad/windows-service-rs"
 license = "MIT OR Apache-2.0"
-edition = "2018"
+edition = "2021"
 
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ environment:
       RUST_VERSION: nightly
     # Testing on oldest supported version of Rust
     - TARGET: x86_64-pc-windows-msvc
-      RUST_VERSION: 1.48.0
+      RUST_VERSION: 1.56.1
 
 install:
   - ps: >-
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ environment:
       RUST_VERSION: nightly
     # Testing on oldest supported version of Rust
     - TARGET: x86_64-pc-windows-msvc
-      RUST_VERSION: 1.34.0
+      RUST_VERSION: 1.48.0
 
 install:
   - ps: >-
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -181,49 +181,49 @@ pub type Result<T> = std::result::Result<T, Error>;
 #[derive(err_derive::Error, Debug)]
 #[error(no_from)]
 pub enum Error {
-    /// Invalid account name.
-    #[error(display = "Invalid account name")]
-    InvalidAccountName(#[error(source)] widestring::NulError<u16>),
+    /// Account name contains a nul byte.
+    #[error(display = "Account name contains a nul byte")]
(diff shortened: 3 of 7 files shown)


faern:
The CI does not pass. Because the MSRV has been increased in `widestring 0.5`. And also, we publicly export types from `widestring`. See the `Error` enum. This makes it a breaking change to upgrade the dependency to a non-semver compatible version.

We should probably not expose the `NulError` in this crate to begin with! It limits us in situations like this.

So we can't merge this PR as is. I think the action plan should be:

1. Redesign the error type in some way to not expose types from `widestring`. And find out if we expose `widestring` types somewhere else in this library.
2. Upgrade this crate to edition 2021. Because that's what's `widestring 0.6` is going to do anyway. Better stay up to date.
3. Release `windows-service 0.5` as a breaking change.

vthib:
Indeed, not exposing widestring on the public API would be beneficial.
I will try doing those changes when I have some time

vthib:
I've done the changes you suggested:
- Addition of a `NulError` that replaces the error from widestring we re-exported on the public API.
- Update of widestring
- Update to edition 2021

Let me know if that looks ok to you

vthib:
ping @faern :)

pronebird (approved):
That actually looks fine.

> Reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/66#-:-N-hFSsO-_mgI1q_QpL-:bl82h07)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (changes requested):
> Reviewed 3 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/66)* status: all files reviewed, 2 unresolved discussions (waiting on @vthib)

___
*[src/error.rs, line 88 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/66#-N019hS92DFbzSSET6yb:-N019hS92DFbzSSET6yc:b2zx5jw) ([raw file](https://github.com/mullvad/windows-service-rs/blob/7b347d88c9721a188d5dff8f3759f097894fbd88/src/error.rs#L88)):*
> ```XML
> /// This error contains the position of the nul value, as well as the faulty string.
> #[derive(Debug, Clone, PartialEq, Eq)]
> pub struct NulError(usize, Option<Vec<u16>>);
> ```

This is fine for moving along with this PR and getting something out there. But when I said we should not expose third party errors, vendoring is not exactly what I had in mind. It would not be feasible to vendor every third party error and just do internal conversion. I don't think that scales in the long run. But it's fine here this time. Because I don't have a clear idea of how to solve it otherwise, and it's just a single type this time.

I think very few need to check the actual `NulError` source error of for example "Invalid display name" error. We could just remove the "source" from that error and make the message "Display name has null byte" and the same for all the others. I don't think that would remove needed functionality for anyone, but I don't know.
___
*[src/error.rs, line 111 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/66#-N01APeW6FDASeQsghYY:-N01APeW6FDASeQsghYZ:b-cw9ayc) ([raw file](https://github.com/mullvad/windows-service-rs/blob/7b347d88c9721a188d5dff8f3759f097894fbd88/src/error.rs#L111)):*
> ```XML
> impl std::error::Error for NulError {
>     fn description(&self) -> &str {
>         "nul value found"
> ```

`Error::description` is deprecated. Please don't implement it.
___
*[src/error.rs, line 115 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/66#-N01AamhAE_PZGdmdmiI:-N01AamhAE_PZGdmdmiJ:b-3j23xl) ([raw file](https://github.com/mullvad/windows-service-rs/blob/7b347d88c9721a188d5dff8f3759f097894fbd88/src/error.rs#L115)):*
> ```XML
> }
> 
> impl From<widestring::error::ContainsNul<u16>> for NulError {
> ```

This still makes `widestring` a part of the public API of this crate. Since `NulError` is public this `From` impl is also public. So anyone can call `windows_service::NulError::from(widestring_error)` and that code will break if we internally upgrade `widestring`.

If we need to convert their error into ours it has to be done with a private function.

faern (commented):
Please remove the vendored `NulError`. And adapt our own error variants to say what's actually wrong (contains null bytes, rather than just "invalid")

> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/66#-:-N186v8280kEDa9tzM-L:bpeqqjb)* status: all files reviewed, 2 unresolved discussions (waiting on @vthib)

faern (approved):
> Reviewed 5 of 8 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/66)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #69, opened 2022-05-06
Add API to expose ServiceStatusHandle as_raw_handle

schultetwin1:
This is needed for users who want to pass SERVICE_STATUS_HANDLE into
other Windows APIs such as the RegisterPowerSettingNotification API

https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerpowersettingnotification

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/69)

### DIFF
--- a/src/service_control_handler.rs
+++ b/src/service_control_handler.rs
@@ -1,6 +1,7 @@
 use std::ffi::OsStr;
 use std::io;
 use std::os::raw::c_void;
+use std::os::windows::prelude::{RawHandle, AsRawHandle};
 use widestring::WideCString;
 use windows_sys::Win32::{
     Foundation::{ERROR_CALL_NOT_IMPLEMENTED, NO_ERROR},
@@ -30,6 +31,15 @@ impl ServiceStatusHandle {
         }
     }
 }
+
+impl AsRawHandle for ServiceStatusHandle {
+    /// Get access to the raw handle to use in other Windows APIs
+    fn as_raw_handle(&self) -> RawHandle {
+        self.0 as _
+    }
+}
+
+
 
 // Underlying SERVICE_STATUS_HANDLE is thread safe.
 // See remarks section for more info:


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/69)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @schultetwin1)

___
*[`src/service_control_handler.rs` line 35 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/69#-N3e-hbC3Lb0Bx9GfFCb:-N3e-hbC3Lb0Bx9GfFCc:b-6nhkat) ([raw file](https://github.com/mullvad/windows-service-rs/blob/07765897f7dbc0b95dbe25f9af5a8a9d07dd3309/src/service_control_handler.rs#L35)):*
> ```XML
> impl AsRawHandle for ServiceStatusHandle {
>     /// Get access to the raw handle to use in other Windows APIs
>     fn as_raw_handle(&self) -> RawHandle {
> ```

What's the benefit of implementing `AsRawHandle` specifically? 

We could simply add a getter for the underlying field and return the underlying field, which can then be casted to whatever power management expects. 

Ideally we probably want to support those notifications providing safe abstractions, but that's perhaps the whole different level of complexity than just adding a getter.

faern (commented):
Could you please rebase this on latest `master`? We have changed from `winapi` to `windows-sys` so the types here changed.

> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/69#-:-N3yQ-eZ25j-dk_CO7zb:b-3o8qpc)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`src/service_control_handler.rs` line 35 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/69#-N3e-hbC3Lb0Bx9GfFCb:-N3yRT-A6K7uI2NIhB26:bn3zjru) ([raw file](https://github.com/mullvad/windows-service-rs/blob/07765897f7dbc0b95dbe25f9af5a8a9d07dd3309/src/service_control_handler.rs#L35)):*

What's the benefit of implementing `AsRawHandle` specifically? 

We could simply add a getter for the underlying field and return the underlying field, which can then be casted to whatever power management expects. 

Ideally we probably want to support those notifications providing safe abstractions, but that's perhaps the whole different level of complexity than just adding a getter.
</blockquote></details>

I think there are benefits with implementing standard traits for common operations (converting to raw handle).

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/69)* status: 0 of 1 files reviewed, all discussions resolved

___
*[`src/service_control_handler.rs` line 35 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/69#-N3e-hbC3Lb0Bx9GfFCb:-N3yVVEV1bPN079Mb2lb:b-z9iv9o) ([raw file](https://github.com/mullvad/windows-service-rs/blob/07765897f7dbc0b95dbe25f9af5a8a9d07dd3309/src/service_control_handler.rs#L35)):*

I think there are benefits with implementing standard traits for common operations (converting to raw handle).
</blockquote></details>

Forgot that the underlying type is `winsvc::SERVICE_STATUS_HANDLE`. :lgtm:

schultetwin1:
> Could you please rebase this on latest master? We have changed from winapi to windows-sys so the types here changed.

Done! Thankfully this change does not touch any types in `winapi` or `windows-sys` so there was no conflict.

pronebird (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/69)* status: :shipit: complete! all files reviewed, all discussions resolved

faern:
Merged. I changed the import to use the `io` module instead of the `prelude`. The prelude is just a re-export and mostly supposed to be used as a wildcard import when writing code heavily using most of those types. I also ran rustfmt as there was a tiny discrepancy.

Thanks for the PR! I hope this will help you use `windows-service`.

schultetwin1:
Not a problem, thank you!

PR #71, opened 2022-06-02
Migrate to `windows-sys`

### LINKED ISSUE #68: Migrate to `windows`/`-sys` crate
`winapi` is a community-maintained crate for windows bindings that has received only a single update in the last year. Microsoft has officially released their own set of bindings providing both a raw low-level crate, `windows-sys`, and a higherish-level crate (with some helper traits and etc), `windows`.

As the general Rust ecosystem is migrating to the new windows crates, `windows-service` should follow. It would be preferable to use the `windows-sys` crate over the `windows` crate to cut down compile times.

yescallop:
Closes #68. This is ready for review now. @faern

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Breaking: Change `winapi` dependency to `windows-sys`. This is a breaking change since some of the low level Windows types are exposed in the public API of this library.
 
 
 ## [0.4.0] - 2021-08-12
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,18 @@ default-target = "x86_64-pc-windows-msvc"
 
 [target.'cfg(windows)'.dependencies]
 bitflags = "1.3"
-err-derive = "0.3.0"
-winapi = { version = "0.3.9", features = ["dbt", "std", "winbase", "winerror", "winsvc"] }
-widestring = "0.5.0"
+err-derive = "0.3.1"
+widestring = "0.5.1"
+
+[target.'cfg(windows)'.dependencies.windows-sys]
+version = "0.36.1"
+features = [
+    "Win32_Foundation",
+    "Win32_Security",
+    "Win32_System_Power",
+    "Win32_System_RemoteDesktop",
+    "Win32_System_Services",
+    "Win32_System_SystemServices",
+    "Win32_System_WindowsProgramming",
+    "Win32_UI_WindowsAndMessaging",
+]
--- a/src/double_nul_terminated.rs
+++ b/src/double_nul_terminated.rs
@@ -1,6 +1,6 @@
 use std::ffi::{OsStr, OsString};
 use widestring::{NulError, WideCStr, WideCString, WideString};
-use winapi::shared::ntdef::LPWSTR;
+use windows_sys::core::PWSTR;
 
 /// A helper to join a collection of `OsStr` into a nul-separated `WideString` ending with two nul
 /// wide characters.
@@ -35,7 +35,7 @@ pub fn from_vec(source: &[impl AsRef<OsStr>]) -> Result<Option<WideString>, NulE
 ///
 /// Output:
 /// ["hello", "world"]
-pub unsafe fn parse_str_ptr(double_nul_terminated_string: LPWSTR) -> Vec<OsString> {
+pub unsafe fn parse_str_ptr(double_nul_terminated_string: PWSTR) -> Vec<OsString> {
     let mut results: Vec<OsString> = Vec::new();
 
     if !double_nul_terminated_string.is_null() {
--- a/src/sc_handle.rs
+++ b/src/sc_handle.rs
@@ -1,21 +1,21 @@
-use winapi::um::winsvc;
+use windows_sys::Win32::{Security, System::Services};
 
-/// A handle holder that wraps a low level [`winsvc::SC_HANDLE`].
-pub(crate) struct ScHandle(winsvc::SC_HANDLE);
+/// A handle holder that wraps a low level [`Security::SC_HANDLE`].
+pub(crate) struct ScHandle(Security::SC_HANDLE);
 
 impl ScHandle {
-    pub(crate) unsafe fn new(handle: winsvc::SC_HANDLE) -> Self {
+    pub(crate) unsafe fn new(handle: Security::SC_HANDLE) -> Self {
         ScHandle(handle)
     }
 
-    /// Returns underlying [`winsvc::SC_HANDLE`].
-    pub(crate) fn raw_handle(&self) -> winsvc::SC_HANDLE {
+    /// Returns underlying [`Security::SC_HANDLE`].
+    pub(crate) fn raw_handle(&self) -> Security::SC_HANDLE {
         self.0
     }
 }
 
 impl Drop for ScHandle {
     fn drop(&mut self) {
-        unsafe { winsvc::CloseServiceHandle(self.0) };
+        unsafe { Services::CloseServiceHandle(self.0) };
     }
 }
--- a/src/service.rs
+++ b/src/service.rs
@@ -9,11 +9,14 @@ use std::time::Duration;
 use std::{io, mem};
 
 use widestring::{NulError, WideCStr, WideCString, WideString};
-use winapi::shared::guiddef::{IsEqualGUID, GUID};
-use winapi::shared::minwindef::DWORD;
-use winapi::shared::winerror::{ERROR_SERVICE_SPECIFIC_ERROR, NO_ERROR};
-use winapi::um::winbase::INFINITE;
-use winapi::um::{dbt, winnt, winsvc, winuser};
+use windows_sys::{
+    core::GUID,
+    Win32::{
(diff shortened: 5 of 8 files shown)


faern:
Cool. I'll take a closer look! Ping, @dlon, could you also please give this a review maybe?

faern (changes requested):
This is a breaking change. So we will need to release version `0.5.0` as the next version if this is merged. Just noting that.

Can you please add this change to the changelog. That we are changing from `winapi` to `windows-sys` and that it's a breaking change since some of the low level Windows types are exposed in the public API of this library.

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-:-N3cV6zl6jQWPEMIYmNB:b-h349gx)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @yescallop)

___
*[`src/service.rs` line 194 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3ccmZk25b-UFdDWwQu:-N3ccmZk25b-UFdDWwQv:bkt72hy) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5c9abd3d86a8a35d158807661f215c04661369dd/src/service.rs#L194)):*
> ```XML
>             x if x == ServiceActionType::Restart.to_raw() => Ok(ServiceActionType::Restart),
>             x if x == ServiceActionType::RunCommand.to_raw() => Ok(ServiceActionType::RunCommand),
>             _ => Err(ParseRawError::InvalidInteger(raw as u32)),
> ```

This will print the wrong number if `raw` is negative, which it could be. If we now have to deal with both `i32`s and `u32`s when parsing raw values I think `ParseRawError` should have an extra variant, to properly deal with each case. I don't think the error types we have that expose all the internal details of an error is ideal, but I also don't think this PR should deal with refactoring that. Just add another variant for now.
___
*[`src/service.rs` line 698 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3cdlEiCmotkK-cEQCM:-N3cdlEiCmotkK-cEQCN:b-7i746c) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5c9abd3d86a8a35d158807661f215c04661369dd/src/service.rs#L698)):*
> ```XML
> 
> fn is_equal_guid(a: &GUID, b: &GUID) -> bool {
>     a.data1 == b.data1 && a.data2 == b.data2 && a.data3 == b.data3 && a.data4 == b.data4
> ```

Feels like the equality check should be implemented for `GUID` directly. But since it's not I guess this has to be here :)

I actually created a PR for this. Maybe it's rejected for some reason. But we'll see: https://github.com/microsoft/windows-rs/pull/1798. Maybe we'll know during the day what the status of that PR will be. But maybe you could add a comment to this function with a link to the PR and say that this can be removed if it's merged and published.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @yescallop)

___
*[`src/service.rs` line 194 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3ccmZk25b-UFdDWwQu:-N3cqV2q0Z3KsjgE3_X7:b-3k5409) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5c9abd3d86a8a35d158807661f215c04661369dd/src/service.rs#L194)):*

This will print the wrong number if `raw` is negative, which it could be. If we now have to deal with both `i32`s and `u32`s when parsing raw values I think `ParseRawError` should have an extra variant, to properly deal with each case. I don't think the error types we have that expose all the internal details of an error is ideal, but I also don't think this PR should deal with refactoring that. Just add another variant for now.
</blockquote></details>

I actually think that many of these raw types are `i32` and not `u32` so we shall perhaps switch `InvalidInteger` to accept `i32` given that in other places it's also `i32`, which I believe is the case since integer is a default type in C for many of these definitions.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @yescallop)

___
*[`src/service.rs` line 194 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3ccmZk25b-UFdDWwQu:-N3ctDua--_Hkoj5REt_:b7b7kfz) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5c9abd3d86a8a35d158807661f215c04661369dd/src/service.rs#L194)):*

You might be correct. I dunno. But I checked the code and even after this PR there are many places that has `raw` as a `u32`, so I'm not sure. They would then need to be casted the other way, which is also suboptimal.
</blockquote></details>

Actually I am gonna retract my last statement. After further inspection it seems that it can either be `i32` or `u32`. So probably having two separate error variants as you suggested is the better way to tackle this.

yescallop (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[`src/service.rs` line 194 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3ccmZk25b-UFdDWwQu:-N3dHBwG9yGBaIso3_pC:b-dcf2uv) ([raw file](https://github.com/mullvad/windows-service-rs/blob/5c9abd3d86a8a35d158807661f215c04661369dd/src/service.rs#L194)):*

Actually I am gonna retract my last statement. After further inspection it seems that it can either be `i32` or `u32`. So probably having two separate error variants as you suggested is the better way to tackle this.
</blockquote></details>

Added `InvalidIntegerSigned` variant on `ParseRawError`.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71)* status: all files reviewed, 2 unresolved discussions (waiting on @yescallop)

___
*[`src/service.rs` line 698 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3dTCJuBVr4CnA3l9Ua:-N3dTCJuBVr4CnA3l9Ub:b-nsqvif) ([raw file](https://github.com/mullvad/windows-service-rs/blob/ef4aef021c58985beac260e69868bbaa9c6f709d/src/service.rs#L698)):*
> ```XML
> 
> // FIXME: Remove this function if microsoft/windows-rs#1798 gets merged and published.
> fn is_equal_guid(a: &GUID, b: &GUID) -> bool {
> ```

`windows-rs` does implement `PartialEq + Eq` for `GUID`. I suppose there's a reason for preferring `-sys` here, though.
___
*[`src/service.rs` line 1746 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3dWhhp89RE7MdnVs8b:-N3dWhhp89RE7MdnVs8c:b-ek2qmg) ([raw file](https://github.com/mullvad/windows-service-rs/blob/a956b5d8ef6a724bb0522b1d403a6bbc79d8d421/src/service.rs#L1746)):*
> ```XML
> }
> 
> fn string_from_guid(guid: &GUID) -> String {
> ```

Why do we roll our own function here? IIRC, `StringFromIID` or `StringFromGUID2` accomplishes the same thing. (I think one of them leaves out braces.)

dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-:-N3dXoKXAY7cRtjvCM8_:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @yescallop)

yescallop (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/71)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`src/service.rs` line 698 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3dTCJuBVr4CnA3l9Ua:-N3daGbE0JCUrwh-_6pU:bnqlpxi) ([raw file](https://github.com/mullvad/windows-service-rs/blob/ef4aef021c58985beac260e69868bbaa9c6f709d/src/service.rs#L698)):*
<details><summary><i>Previously, dlon (David Lönnhager) wrote…</i></summary><blockquote>

`windows-rs` does implement `PartialEq + Eq` for `GUID`. I suppose there's a reason for preferring `-sys` here, though.
</blockquote></details>

The main concern here is compile time, if I understand correctly.
___
*[`src/service.rs` line 1746 at r2](https://reviewable.io/reviews/mullvad/windows-service-rs/71#-N3dWhhp89RE7MdnVs8b:-N3dYSF8AIv3h4ugbDHc:bq1pxyt) ([raw file](https://github.com/mullvad/windows-service-rs/blob/a956b5d8ef6a724bb0522b1d403a6bbc79d8d421/src/service.rs#L1746)):*
<details><summary><i>Previously, dlon (David Lönnhager) wrote…</i></summary><blockquote>

Why do we roll our own function here? IIRC, `StringFromIID` or `StringFromGUID2` accomplishes the same thing. (I think one of them leaves out braces.)
</blockquote></details>

That may be because we're converting the GUID to a `String`. Using `StringFromGUID2` would complicate things as it requires unsafe code.

faern:
There were some slight inconsistencies with line endings being a mix of Windows and unix in these commits. But I realized we had some of those even before this PR. So I'll do a separate commit after where I just fix all line endings.

Thanks for this PR!

PR #73, opened 2022-06-22
build(deps): update `widestring` 0.5.1 -> 1

ErichDonGubler:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/73)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Changed
 - Breaking: Change `winapi` dependency to `windows-sys`. This is a breaking change since some of the low level Windows types are exposed in the public API of this library.
+- Breaking: Update `widestring` dependency from version 0.5.1 to 1. This is a breaking change since `NulError` (now `widestring::error::ContainsNul`) is exposed in the public API of this library.
 
 
 ## [0.4.0] - 2021-08-12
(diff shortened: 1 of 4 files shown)


faern:
Nice! Thanks for this contribution. I'll review it as soon as I have time. But if they went "stable" with a 1.0 release it really sounds like we should adopt it. And given that it's a breaking change, and we have a breaking release coming up, it sounds great to have it merged before we do that release.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Changed
 - Breaking: Change `winapi` dependency to `windows-sys`. This is a breaking change since some of the low level Windows types are exposed in the public API of this library.
+- Breaking: Update `widestring` dependency from version 0.5.1 to 1. This is a breaking change since `NulError` (now `widestring::error::ContainsNul`) is exposed in the public API of this library.
```
vthib, line 10: This is not exactly true. Since https://github.com/mullvad/windows-service-rs/pull/66 widestring is not part of the public API anymore.

So since 0.4.0, there is indeed a breaking change, but it is to vendor this NulError and no longer expose widestring in the public API (see the previously mentioned PR), it isn't the update of widestring to 1.0

faern: Yes and no. Since `windows-service 0.4.0` upgrading `widestring` and changing the error type, is a breaking change. So it should be mentioned in the changelog. BUT the upgrade from 0.5 -> 1.0 is not the breaking part. But I can fix that. I need to polish the changelog anyway.

faern (approved):
Looks good. Thanks for the contribution!

ErichDonGubler:
CC @retep998, @seritools, @Xaeroxe.☝🏻

PR #74, opened 2022-07-19
Improve CI and do some cleanups

faern:
The CI config for this crate is very old. We have migrated everything else to Github Actions. So let's do that for this one as well.

As part of adding formatting and linting I fixed everything Clippy was complaining about. And I moved our formatting to the stable Rust formatter instead of nightly.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/74)


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up MinGW
+        if: matrix.target == 'i686-pc-windows-gnu' || matrix.target == 'x86_64-pc-windows-gnu'
+        uses: egor-tensin/setup-mingw@v2
+        with:
+          platform: ${{ matrix.platform }}
```
faern, line 35: I noticed that the CI works without this parameter also. so for some reason `i686-pc-windows-gnu` builds when this is unset. The documentation for this action says that the default is x64, but I guess it installs both toolchains by default then. I guess it does not hurt to specify it more explicitly? Except a bit more verbose CI config... I'm a bit afraid we test the wrong thing without it :thinking:

dlon (approved):
> Reviewed 7 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/74)* status: all files reviewed, 1 unresolved discussion

PR #75, opened 2022-07-20
Revert Service::start back to using generics over impl trait

### LINKED ISSUE #51: Cannot start service with empty arguments
When passing an empty array to `Service::start` Rust needs to be told the type as it cannot infer the type of an empty array. Before I could do `service.start::<OsString>(&[])?;`. In 0.3.0 `Service::start` was changed to use `impl AsRef<OsStr>` instead of an explicit type parameter which means I cannot tell Rust what the type is. Instead I have to do this:
```Rust
let arguments: &[OsString] = &[];
service.start(arguments)?;
```
Please reconsider.

faern:
Reverting back to how the signature looked in version 0.2.
Allows calling it without any arguments in a more ergonomic way.
`service.start::<OsString>(&[])` should now work again.

I don't see any downsides of this. The impl trait way of writing it does not really have any benefits, except it does not litter functions with generics. But here it's just a single generic, and it has an actual use case. Even if there are ways of calling `Service::start` without arguments anyway, it is more ergonomic with generics.

Fixes #51

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -1376,7 +1376,7 @@ impl Service {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn start(&self, service_arguments: &[impl AsRef<OsStr>]) -> crate::Result<()> {
+    pub fn start<S: AsRef<OsStr>>(&self, service_arguments: &[S]) -> crate::Result<()> {
         let wide_service_arguments = service_arguments
             .iter()
             .map(|s| WideCString::from_os_str(s).map_err(|_| Error::StartArgumentHasNulByte))


pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/75)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/75#-:-N7PciNN1F2ezVic-vxS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #79, opened 2022-08-10
ServiceManager::Add get_service_key_name()

Guiguiprim:
Add this helper function to help finding services from there display names.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/79)

### DIFF
--- a/src/service_manager.rs
+++ b/src/service_manager.rs
@@ -1,4 +1,4 @@
-use std::ffi::OsStr;
+use std::ffi::{OsStr, OsString};
 use std::{io, ptr};
 
 use widestring::WideCString;
@@ -206,6 +206,47 @@ impl ServiceManager {
             Err(Error::Winapi(io::Error::last_os_error()))
         } else {
             Ok(Service::new(unsafe { ScHandle::new(service_handle) }))
+        }
+    }
+
+    /// Search the service name associated with a service display name.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - A service display name.
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// use windows_service::service_manager::{ServiceManager, ServiceManagerAccess};
+    ///
+    /// # fn main() -> windows_service::Result<()> {
+    /// let manager = ServiceManager::local_computer(None::<&str>, ServiceManagerAccess::CONNECT)?;
+    /// let my_service_name = manager.get_service_key_name("My Service Display Name")?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn get_service_key_name(&self, display_name: impl AsRef<OsStr>) -> Result<OsString> {
+        let service_display_name =
+            WideCString::from_os_str(display_name).map_err(|_| Error::DisplayNameHasNulByte)?;
+        // As per docs, the maximum size of data buffer used by GetServiceKeyNameW is 4k
+        let mut size = 4u32 * 1024;
+        let mut buffer = vec![0u16; size as usize];
+        let result = unsafe {
+            Services::GetServiceKeyNameW(
+                self.manager_handle.raw_handle(),
+                service_display_name.as_ptr(),
+                buffer.as_mut_ptr(),
+                &mut size as _,
+            )
+        };
+
+        if result == 0 {
+            Err(Error::Winapi(io::Error::last_os_error()))
+        } else {
+            let service_name = WideCString::from_vec(&buffer[..size as usize])
+                .map_err(|_| Error::ServiceNameHasNulByte)?;
+            Ok(service_name.to_os_string())
         }
     }
 }


src/service_manager.rs
```diff
@@ -208,4 +209,41 @@ impl ServiceManager {
...
+    pub fn get_service_key_name(&self, display_name: impl AsRef<OsStr>) -> Result<OsString> {
+        let service_display_name =
+            WideCString::from_os_str(display_name).map_err(|_| Error::DisplayNameHasNulByte)?;
+        let mut size = 0u32;
+        let res = unsafe {
+            Services::GetServiceKeyNameW(
+                self.manager_handle.raw_handle(),
+                service_display_name.as_ptr(),
+                std::ptr::null_mut(),
+                &mut size as _,
```
faern, line 223: Instead of probing for the buffer size we usually just allocate the maximum buffer size directly and get away with one syscall. Because a small allocation is definitely faster than a syscall. Unless there is some syscall fast path here I'm not aware of.

See this method for reference: https://github.com/mullvad/windows-service-rs/blob/24bd91a84529a315277b7096a52d67057a394edb/src/service.rs#L1465-L1487

According to [the docs](https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-getservicekeynamew) the max size should be 4k.

What do you think about this approach here, and in general?

src/service_manager.rs
```diff
@@ -208,4 +209,41 @@ impl ServiceManager {
...
+            )
+        };
+
+        if res == 0 && unsafe { GetLastError() } != ERROR_INSUFFICIENT_BUFFER {
+            return Err(Error::Winapi(io::Error::last_os_error()));
```
faern, line 228: This effectively calls `GetLastError` twice. I think we should get this down to a single time one way or another. You can store away the error and use `Error::from_raw_os_error` for example. But given the feedback above, this code might no longer be needed at all.

src/service_manager.rs
```diff
@@ -208,4 +209,41 @@ impl ServiceManager {
...
+        let res = unsafe {
+            Services::GetServiceKeyNameW(
+                self.manager_handle.raw_handle(),
+                service_display_name.as_ptr(),
+                buffer.as_mut_ptr(),
+                &mut size as _,
+            )
+        };
+
+        if res == 0 {
```
faern, line 241: Nitpick. But please call the variable `result`. Even if it's a commonly used abbreviation it does not really serve any purpose other than making the code harder to read. Adding `ult` to the variable name will not split any code in more lines or anything.

src/service_manager.rs
Guiguiprim: I'm used to allocate buffers of the needed space, but I guess this approach works fine (as long as the max size is reasonable for a default value).

Done

src/service_manager.rs
Guiguiprim: Removed

src/service_manager.rs
Guiguiprim: Done

src/service_manager.rs
```diff
@@ -208,4 +208,28 @@ impl ServiceManager {
             Ok(Service::new(unsafe { ScHandle::new(service_handle) }))
         }
     }
+
+    /// Retrieves the service name of the specified service.
```
faern, line 212: I know this is just a verbatim copy of the MS docs.. But I think we can do better. I personally find it very hard to know when and how to use this from just reading this docs and seeing the function signature. See `create_service` and `open_service` above for inspiration. Please write more detailed docs for this method.

src/service_manager.rs
```diff
@@ -208,4 +208,28 @@ impl ServiceManager {
...
+    pub fn get_service_key_name(&self, display_name: impl AsRef<OsStr>) -> Result<OsString> {
+        let service_display_name =
+            WideCString::from_os_str(display_name).map_err(|_| Error::DisplayNameHasNulByte)?;
+        let mut size = 4u32 * 1024;
```
faern, line 216: Please leave a comment describing why this magic number was chosen. Like here https://github.com/mullvad/windows-service-rs/blob/24bd91a84529a315277b7096a52d67057a394edb/src/service.rs#L1466

faern (changes requested):

src/service_manager.rs
Guiguiprim: Done

src/service_manager.rs
Guiguiprim: Done

Guiguiprim:
Sorry, I forgot about this (using my fork...)

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/79)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @Guiguiprim)

___
*[`src/service_manager.rs` line 234 at r3](https://reviewable.io/reviews/mullvad/windows-service-rs/79#-NEHb8YJ1Jn3Gtj9kjW4:-NEHb8YJ1Jn3Gtj9kjW5:b-pstok8) ([raw file](https://github.com/mullvad/windows-service-rs/blob/b896664a0063d41208f59791a024dc8997c03311/src/service_manager.rs#L234)):*
> ```XML
>         // As per docs, the maximum size of data buffer used by GetServiceKeyNameW is 4k
>         let mut size = 4u32 * 1024;
>         let mut buffer = vec![0u16; size as usize];
> ```

The docs specify the max size as 4k *bytes*. This vector is 2*4k bytes since the elements are `u16`s.

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/79)* status: 0 of 1 files reviewed, 7 unresolved discussions (waiting on @faern and @Guiguiprim)

___
*[`src/service_manager.rs` line 212 at r3](https://reviewable.io/reviews/mullvad/windows-service-rs/79#-NEKV3WT1YUtD2YU7YGh:-NEKV3WT1YUtD2YU7YGi:bdpvywz) ([raw file](https://github.com/mullvad/windows-service-rs/blob/b896664a0063d41208f59791a024dc8997c03311/src/service_manager.rs#L212)):*
> ```XML
>     }
> 
>     /// Search the service name associated with a service display name.
> ```

Could this be changed to something like "Return the name of a service given its display name."?
___
*[`src/service_manager.rs` line 229 at r3](https://reviewable.io/reviews/mullvad/windows-service-rs/79#-NEKTzkQ9HK3468Oy4-F:-NEI5Pgj0GIEUd0njPgK:b-aifavw) ([raw file](https://github.com/mullvad/windows-service-rs/blob/b896664a0063d41208f59791a024dc8997c03311/src/service_manager.rs#L229)):*
> ```XML
>     /// # }
>     /// ```
>     pub fn get_service_key_name(&self, display_name: impl AsRef<OsStr>) -> Result<OsString> {
> ```

Could this be renamed to something like `service_name_from_display_name` instead? The original function name is not very good.
___
*[`src/service_manager.rs` line 240 at r3](https://reviewable.io/reviews/mullvad/windows-service-rs/79#-NEI0oiz-DgUZKleimya:-NEI0oj-4TPkohwOi6Tp:bkpz8bv) ([raw file](https://github.com/mullvad/windows-service-rs/blob/b896664a0063d41208f59791a024dc8997c03311/src/service_manager.rs#L240)):*
> ```XML
>                 service_display_name.as_ptr(),
>                 buffer.as_mut_ptr(),
>                 &mut size as _,
> ```

`as _` is redundant.
___
*[`src/service_manager.rs` line 247 at r3](https://reviewable.io/reviews/mullvad/windows-service-rs/79#-NEHuIio6HSiw_000ypi:-NEHuIio6HSiw_000ypj:b-lfxp8l) ([raw file](https://github.com/mullvad/windows-service-rs/blob/b896664a0063d41208f59791a024dc8997c03311/src/service_manager.rs#L247)):*
> ```XML
>             Err(Error::Winapi(io::Error::last_os_error()))
>         } else {
>             let service_name = WideCString::from_vec(&buffer[..size as usize])
> ```

It would be slightly nicer to use [`OsStringExt::from_wide`](https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStringExt.html) here:

```rust
Ok(OsString::from_wide(&buffer[..size as usize]))
```

PR #87, opened 2023-02-04
Bug fixes, features and some public API changes

citreae535:
### 1. Breaking: change `Service.delete` to borrow rather than consume `self`
The underlying win32 API call merely marks the service for deletion. It can be called even if the service is not stopped and will not invalidate the caller's open handle to it. Making the function not consuming `self` allows the caller to decide when to close the handle. Docs and examples are updated to reflect this change.
### 2. Refactor `double_nul_terminated` implementation
- Rename `from_vec` to `from_slice` since it takes a slice, not a `Vec`.
- Allocate `WideString` in `from_slice` with pre-calculated capacity.
- Add unit tests for `from_slice`.
### 3. Bump `windows-sys` to `0.45`
An additional feature `Win32_Storage_FileSystem` is pulled in just for the flag `FileSystem::DELETE`. This does not feel right, but hard-coding it ourselves feels worse.

### 4. Add public method `Service.set_delayed_auto_start`
This function configures if an auto-start service should be delayed.
### 5. Add public method `Error.is_os_error`
This function checks if an error from win32 API call matches a specific OS error code.

## Public API changes
These three are more opinionated public API changes based on my own needs, If these do not feel right for you, I'll revert them.
### 6. Breaking: remove `ServiceDependency`
- It does not seem to do anything useful. The '+' prefix of a group dependency has to be passed to the win32 API. An `OsString` will suffice here since it is up to users to distinguish between service and group dependencies.
- If users want to check if a dependency is a group, they can simply check if the `OsString` starts with '+'.
### 7. Breaking: change `ServiceInfo` to a borrowed type
This change allows the caller to retain individual ownership of the fields of `ServiceInfo` without having to clone them.
### 8. Possibly breaking: replace `err-derive` with `thiserror`
  - Both crates derive `std::error::Error` for an enum with slightly different syntax. For this crate it is a drop-in replacement and unlikely to break anything.
  - `thiserror` has some extra features and are better maintained.
  - Because `thiserror` is vastly more popular (76,032,688 total downloads and 8400 dependent crates compared to `err-derive`'s 2,175,409 and 71), it is very likely that many users of this crate already depend on `thiserror` and will have less dependencies if  this switch is made. (In fact, crates.io shows that this crate is the most downloaded dependent of `err-derive`)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/87)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ default-target = "x86_64-pc-windows-msvc"
 
 [target.'cfg(windows)'.dependencies]
 bitflags = "1.3"
-err-derive = "0.3.1"
+thiserror = "1"
 widestring = "1"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,11 @@ err-derive = "0.3.1"
 widestring = "1"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.36.1"
+version = "0.45"
 features = [
     "Win32_Foundation",
     "Win32_Security",
+    "Win32_Storage_FileSystem",
     "Win32_System_Power",
     "Win32_System_RemoteDesktop",
     "Win32_System_Services",
--- a/examples/install_service.rs
+++ b/examples/install_service.rs
@@ -1,6 +1,6 @@
 #[cfg(windows)]
 fn main() -> windows_service::Result<()> {
-    use std::ffi::OsString;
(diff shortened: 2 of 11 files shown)


pronebird:
Hi,

Thanks for your efforts. However, your PR introduces a lot of changes, some of them are good, others are breaking or subject to discussion. It would be much simpler to review your changes if they were somehow isolated one from the other.

citreae535:
Close in favor of separate PRs

PR #89, opened 2023-02-07
Add `Error.is_os_error`

citreae535:
This function checks if an error from win32 API call matches a specific OS error code. This makes it easier for users to check specific API errors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/89)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -269,12 +269,10 @@ impl Error {
     ///
     /// [official docs]: https://learn.microsoft.com/en-us/windows/win32/api/winsvc/#functions
     pub fn is_os_error(&self, error_code: i32) -> bool {
-        if let Self::Winapi(e) = self {
-            if e.raw_os_error() == Some(error_code) {
-                return true;
-            }
+        match self {
+            Self::Winapi(e) => e.raw_os_error() == Some(error_code),
+            _ => false,
         }
-        false
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -258,6 +258,26 @@ pub enum Error {
     Winapi(#[error(source)] std::io::Error),
 }
 
+impl Error {
+    /// Return if an error from win32 API call matches the specific OS error code.
+    ///
+    /// To find the names of OS error codes that a specific win32 API call may return,
+    /// browse the [official docs]. They are also defined in `windows_sys` crate.
+    ///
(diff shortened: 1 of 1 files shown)


pronebird (changes requested):
Nice addition. A few leftovers in this PR just like in all others.

> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/89#-:-NNkh9cVDBUXBZc37TGT:b-t9mipq)* status: 1 of 7 files reviewed, 1 unresolved discussion (waiting on @citreae535)

___
*[`src/lib.rs` line 272 at r1](https://reviewable.io/reviews/mullvad/windows-service-rs/89#-NNkeHX1EGWSRXbotYJ8:-NNkeHX1EGWSRXbotYJ9:bz4heol) ([raw file](https://github.com/mullvad/windows-service-rs/blob/c4fbc76dce8f2ee53809ea57ba1e13097fe09b14/src/lib.rs#L272)):*
> ```XML
>     /// [official docs]: https://learn.microsoft.com/en-us/windows/win32/api/winsvc/#functions
>     pub fn is_os_error(&self, error_code: i32) -> bool {
>         if let Self::Winapi(e) = self {
> ```

I think historically we prefer to avoid `return`. This could have been written as a match statement instead, i.e:

```rs
match self {
    Self::Winapi(e) => e.raw_os_error() == Some(error_code),
    _ => false
}
```

faern:
I'm not against this PR. But I also want to note that this crate is supposed to be a high level abstraction for Windows services, and a user of the library should preferably not need to deal with OS error codes or other `windows-sys` stuff if possible.

citreae535:
Hmm, I reconsidered it and have to agree that most users won't need this change. Better close it then.

pronebird:
Currently we do not translate errors returned from syscalls into meaningful errors. Instead we return `Winapi(io::Error)` variant so the one has to parse `io::Error` to find the cause of error. 

We could of course introduce new error types and provide error variants for common error codes but we would almost always have to have a `Other(io::Error)` variant since we do not control Windows API. I am not quite sure that beefing up a single `Error` type would be the way to go, on the other side having dedicated errors types would mean that we'd have to have duplicate error variants like those that we currently have in the `Error` type.

Anyhow, reinventing that would be a big task.

PR #90, opened 2023-02-07
Breaking: change `Service.delete` to borrow rather than consume `self`

citreae535:
The underlying win32 API call merely marks the service for deletion. It can be called even if the service is not stopped and will not invalidate the caller's open handle to it. Making the function not consuming self allows the caller to decide when to close the handle. Docs and examples are updated to show the correct behavior of this API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/90)

### DIFF
--- a/examples/uninstall_service.rs
+++ b/examples/uninstall_service.rs
@@ -1,10 +1,10 @@
 #[cfg(windows)]
 fn main() -> windows_service::Result<()> {
-    use std::{thread, time::Duration};
     use windows_service::{
         service::{ServiceAccess, ServiceState},
         service_manager::{ServiceManager, ServiceManagerAccess},
     };
+    use windows_sys::Win32::Foundation::ERROR_SERVICE_DOES_NOT_EXIST;
 
     let manager_access = ServiceManagerAccess::CONNECT;
     let service_manager = ServiceManager::local_computer(None::<&str>, manager_access)?;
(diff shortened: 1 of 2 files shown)


pronebird (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/90)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
I think it makes sense esp. because deletion can fail so it makes no sense to consume the `Service`. I am not sure there was any need to go through all examples and change `println!` but I'll leave it up to someone else to review. I am still using `println!` in the old ways and not very used to `{variable}`.

> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/90#-:-NNkLYHTERUtGzV9CYbX:b8gldq)* status: :shipit: complete! all files reviewed, all discussions resolved

citreae535:
Those `println!` are changed because this clippy lint, [uninlined_format_args](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args), shows up as warnings by default in `1.67.0`. It was [changed](https://github.com/rust-lang/rust-clippy/pull/10265) to allow in `1.67.1` but the maintainers said they "should restore this when rust-analyzer support gets better". As long as it is consistent I don't really have a strong opinion on this. Might be a good idea adding a `clippy.toml` if the old way is preferable.

citreae535:
This one can wait for #89 to make use of that error helper function in the example.

citreae535:
Just rebased it. Should be ready to merge.

pronebird:
In the uninstall example we had a call to `thread::sleep()` to let the service stop. Do you think we should keep it to be on the safe side?

pronebird:
@citreae535  previous PRs are already merged so you could perhaps rebase this one and see my other comment above too.

citreae535:
The uninstall example now polls for service stop and deletion with a timeout.

pronebird:
Merged. Good work & thanks for contributions! 👍

Issue #92, opened 2023-02-07
Some proposals for public API changes


citreae535:
### Remove `ServiceDependency`
This abstraction does not seem to do anything useful but some superfluous allocations. We just need to pass whatever callers give us directly to the win32 API and don't need to distinguish between service and group dependencies ourselves. If users want to check if a dependency is a group, they can simply check if it starts with '+'.

### Replace `err-derive` with `thiserror`
Both crates derive `std::error::Error` for an `enum` with slightly different syntax. `thiserror` should be a drop-in replacement and unlikely to break anything. It has some extra features and seems to be better maintained, based on the update time and frequency.

My main motivation for this change is to reduce two functionally similar dependencies to one. `thiserror` is vastly more popluar, therefore users are much more likely to already depend on it rather than `err-derive`. (I just noticed that your VPN client uses `err-derive` but already depended on `thiserror`.)

For comparison, `thiserror` has 76,032,688 total downloads and 8400 dependent crates listed on crates.io when I checked recently. `err-derive` has 2,175,409 total downloads and 71 dependent crates. (This crate happens to be its most downloaded dependent.)

pronebird:
I don't have anything against `ServiceDependency`. I think I was the one who wrote it and I think it's convenient abstraction and does not need to be removed. Micro allocations that it does are negligent on modern computers. It's not 80186 times anymore. It doesn't mean that we should be wasteful, but I think we should draw a line between micro (or even nano) optimizations and real performance bottlenecks vs. convenience and believe it or not, easy to forget what that `+` prefix does. On the other side we maintain a fine line between being merely a wrapper for windows services and being a high level abstraction. We're more on a wrapper side but not entirely without a guilt when comes to abstractions.

Speaking of error handling, [there was an attempt to rewrite it without using external crate](https://github.com/mullvad/windows-service-rs/pull/48). I think it's reasonable and our errors are relative simple and flat and so dropping dependency would be good in my opinion esp. that ancient error chain thing.I don't know why it stalled, 2020 was a mess and I think it should be rebased and revisited.

citreae535:
Well, I had assumed that users of this crate would know what '+' does, but if that's not the case, then yeah, this abstraction is ok.

[...]

PR #94, opened 2023-02-13
Manual error implementation with less variants

citreae535:
As discussed in #92.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ default-target = "x86_64-pc-windows-msvc"
 
 [target.'cfg(windows)'.dependencies]
 bitflags = "1.3"
-err-derive = "0.3.1"
 widestring = "1"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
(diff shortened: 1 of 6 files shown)


pronebird (approved):
Love it. :lgtm:

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/94#-:-NOAiRry5BL9nZmJV5Cp:b-ptxh7y)* status: :shipit: complete! all files reviewed, all discussions resolved

faern:
I'm all for removing the error helper libraries from the dependency tree. I'm not 100% convinced these are the best error variant formats. They still expose the internal details of each variant and it's going to be a breaking change to make any change to the error enum. But it's at least not worse than before. I don't know how people consume this crate actually match on errors, but you might be correct that no one match on the exact arguments having null in them. However, if you were writing a GUI or whatever that allows you to control services, knowing which field had an error in it could potentially help with reporting the error back to the user in a better way. But this might be overthinking it.

We can probably merge this.

pinkisemils (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/94)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #97, opened 2023-02-15
Add `Service.set_preshutdown_timeout`

citreae535:
This function wraps the Win32 API which configures the pre-shutdown timeout value of a service. The timeout is only relevant for services that accept and handle `ServiceControl::Preshutdown` control code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/97)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -1715,6 +1715,27 @@ impl Service {
         }
     }
 
+    /// Set the preshutdown timeout value of the service.
+    ///
+    /// When the system prepares to shutdown, the service control manager will send [`ServiceControl::Preshutdown`]
+    /// to any service that accepts [`ServiceControlAccept::PRESHUTDOWN`] and block shutdown until either the
+    /// service is stopped or the preshutdown timeout has elapsed. The default value is 180 seconds on releases
+    /// prior to Windows 10 build 15063, and 10 seconds afterwards. This value is irrelevant unless the service
+    /// handles [`ServiceControl::Preshutdown`].
+    ///
+    /// Panics if the specified timeout is too large to fit as milliseconds in a `u32`.
+    ///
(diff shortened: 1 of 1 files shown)


pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/97)* status: :shipit: complete! all files reviewed, all discussions resolved

src/service.rs
```diff
@@ -1715,6 +1715,27 @@ impl Service {
...
+    /// service is stopped or the preshutdown timeout has elapsed. The default value is 180 seconds on releases
+    /// prior to Windows 10 build 15063, and 10 seconds afterwards. This value is irrelevant unless the service
+    /// handles [`ServiceControl::Preshutdown`].
+    ///
+    /// Panics if the specified timeout is too large to fit as milliseconds in a `u32`.
+    ///
+    /// Required permission: [`ServiceAccess::CHANGE_CONFIG`].
+    pub fn set_preshutdown_timeout(&self, timeout: Duration) -> crate::Result<()> {
+        let mut timeout = Services::SERVICE_PRESHUTDOWN_INFO {
+            dwPreshutdownTimeout: timeout.as_millis() as u32,
```
pronebird, line 1731: FYI we do something like that in other part of the code. I am not sure if it really adds much clarity on the fact that we expect it to panic on too long `Duration`, but I thought I'd share it:

```rs
u32::try_from(delay.as_millis()).expect("Too long delay")
```

pronebird (approved):
Overall looks sane.

src/service.rs
pronebird: I've addressed that in https://github.com/mullvad/windows-service-rs/pull/98. Thanks for contributing!

PR #109, opened 2023-11-16
deps: update to windows-sys 0.52.0

poliorcetics:
`windows-sys` moved its MSRV to 1.56, which is less or equal to the one in this crate since the
edition for `windows-service` is 2021, so no impact there.

Could we have a release after this is merged ? ~~I don't see any `pub use` of `windows_sys::` stuff
so a point release should be enough.~~ `windows-sys` is part of the public API so it would need a breaking release.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/109)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ bitflags = "2.3"
 widestring = "1"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.48.0"
+version = "0.52.0"
 features = [
     "Win32_Foundation",
     "Win32_Security",


faern:
Thanks for the PR!

`windows-sys` is part of our public API. For example in a bunch of `to_raw` methods, such as [ServiceAction::to_raw](https://docs.rs/windows-service/latest/windows_service/service/struct.ServiceAction.html#method.to_raw) for example. This means we need to bump our version in a breaking way, not just a patch release.

I don't know exactly what changed in the new version of the crate, but I guess it's nice we did not need any code changes to make it work :)

The CI fails, but I don't think it's because of these changes...

faern:
Here is why the CI is failing: https://github.com/egor-tensin/setup-mingw/issues/15

faern:
@poliorcetics We have now fixed the mingw issue. You can rebase this PR and push it to see if it works now.

poliorcetics:
Thanks for the notif, rebased 👍

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/109)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon:
Thanks!

PR #113, opened 2023-12-20
Add `SystemStart` and `BootStart` start types

kkent030315:
Added `SystemStart` (`SERVICE_SYSTEM_START`) and `BootStart` (`SERVICE_BOOT_START`) start types.
This change has been tested locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/113)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -93,6 +93,10 @@ pub enum ServiceStartType {
     OnDemand = Services::SERVICE_DEMAND_START,
     /// Disabled service
     Disabled = Services::SERVICE_DISABLED,
+    /// Start on system startup
+    SystemStart = Services::SERVICE_SYSTEM_START,
+    /// Start on OS boot
+    BootStart = Services::SERVICE_BOOT_START,
 }
 
 impl ServiceStartType {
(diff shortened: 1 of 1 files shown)


pronebird:
Would be great to document that these two new constants are used for drivers to avoid confusion with a regular `SERVICE_AUTO_START`.

See: https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-createservicea

kkent030315:
> Would be great to document that these two new constants are used for drivers to avoid confusion with a regular `SERVICE_AUTO_START`.
> 
> See: https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-createservicea

I've added more precise comments on them to avoid confusing with usermode service start types. :+1:

faern (changes requested):
Sounds like a good change. This will be a breaking change since we add variants to a public enum. I think we should mark this enum as `#[non_exhaustive]` to prevent that in the future.

kkent030315:
> Sounds like a good change. This will be a breaking change since we add variants to a public enum. I think we should mark this enum as `#[non_exhaustive]` to prevent that in the future.

@faern Thank you for the review. I've added the `#[non_exhaustive]` attribute for the backwards compatibility (81331bced5f8d2fe4df071cad2409c859303f46d).

dlon (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/113#-:-Nmf2xq62xZtoRPhTyk-:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @kkent030315)

___
*[`src/service.rs` line 89 at r3](https://reviewable.io/reviews/mullvad/windows-service-rs/113#-Nmf2Lv1-Onc3kqaoEig:-Nmf2Lv1-Onc3kqaoEih:b9ncvmw) ([raw file](https://github.com/mullvad/windows-service-rs/blob/81331bced5f8d2fe4df071cad2409c859303f46d/src/service.rs#L89)):*
> ```XML
> #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
> #[repr(u32)]
> #[non_exhaustive]
> ```

It looks like this actually is exhaustive now. So maybe this attribute is unnecessary.

kkent030315:
@dlon Thank you for the review. I've removed the `#[non_exhaustive]` in d450630322e9c5897855dd10f3040cedffa77a77.

PR #115, opened 2024-01-16
Add some missing flags and expose service handle

NuSkooler:
This adds some of the ALL_ACCESS flags + exposes the raw service handle. I'm personally using this for [SetServiceObjectSecurity](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-setserviceobjectsecurity). 

One could expose a more Rust-like `set_service_object_security()` call, though it could become fairly complex. I'm hesitant to try to do that right away before the library is based off of the official Windows crates.

Thoughts?

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/115)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "windows-service"
-version = "0.6.0"
+version = "0.6.1"
 description = "A crate that provides facilities for management and implementation of windows services"
 readme = "README.md"
 authors = ["Mullvad VPN"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "windows-service"
-version = "0.6.1"
+version = "0.6.2"
 description = "A crate that provides facilities for management and implementation of windows services"
 readme = "README.md"
 authors = ["Mullvad VPN"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "windows-service"
-version = "0.6.2"
+version = "0.6.0"
 description = "A crate that provides facilities for management and implementation of windows services"
 readme = "README.md"
 authors = ["Mullvad VPN"]
--- a/src/service.rs
+++ b/src/service.rs
@@ -1442,6 +1442,11 @@ impl Service {
         Service { service_handle }
     }
 
+    /// Provides access to the underlying system service handle
+    pub fn raw_handle(&self) -> Security::SC_HANDLE {
+        self.service_handle.raw_handle()
+    }
+
     /// Start the service.
     ///
     /// # Example
@@ -1845,10 +1850,6 @@ impl Service {
             self.change_config2(Services::SERVICE_CONFIG_PRESHUTDOWN_INFO, &mut timeout)
                 .map_err(Error::Winapi)
         }
-    }
-
-    pub fn raw_service_handle(&self) -> Security::SC_HANDLE {
-        self.service_handle.raw_handle()
     }
 
     /// Private helper to send the control commands to the system.
--- a/src/service.rs
+++ b/src/service.rs
@@ -14,6 +14,7 @@ use windows_sys::{
     Win32::{
         Foundation::{ERROR_SERVICE_SPECIFIC_ERROR, NO_ERROR},
         Storage::FileSystem,
+        Security,
         System::{Power, RemoteDesktop, Services, SystemServices, Threading::INFINITE},
         UI::WindowsAndMessaging,
     },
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "windows-service"
-version = "0.6.0"
+version = "0.6.2"
```
faern, line 3: Please don't bump the verison in feature PRs. We'll do that prior to release in a manner suitable for the collective changes since last release.

src/service.rs
```diff
@@ -1843,6 +1847,10 @@ impl Service {
         }
     }
 
+    pub fn raw_service_handle(&self) -> Security::SC_HANDLE {
```
faern, line 1850: We have a lot of to-raw exposure in this crate already, and we usually name the methods `to_raw` unless I have missed some places where we are inconsistent(?). Can you please follow the same naming convention?

I also think this type of method should either be close to the top or bottom of methods, not somewhere in the middle. Please look at similar examples and try to be consistent with those (if we have a pattern? But I think we do).

src/service_manager.rs
```diff
@@ -21,6 +21,8 @@ bitflags::bitflags! {
 
         /// Can enumerate services or receive notifications.
         const ENUMERATE_SERVICE = Services::SC_MANAGER_ENUMERATE_SERVICE;
+
+        const ALL_ACCESS = Services::SC_MANAGER_ALL_ACCESS;
```
faern, line 25: Is there any suitable documentation you could add here? The more descriptive the better :)

faern (changes requested):
Thank you for this addition! See my individual comments for some small improvement suggestions.

src/service_manager.rs
NuSkooler: Updated, though I admit it's still a little vague. From MSDN:

> SC_MANAGER_ALL_ACCESS (0xF003F) | Includes STANDARD_RIGHTS_REQUIRED, in addition to all access rights in this table.

src/service.rs
NuSkooler: I found a number of `to_raw`, but in this case, I changed it to mirror the underlying `raw_handle`. If you'd like `to_raw`, I can of course make that change!

NuSkooler:
> Thank you for this addition! See my individual comments for some small improvement suggestions.

Thanks for the quick response! Made some quick updates, happy to adjust whatever else is needed!

src/service_manager.rs
pronebird: Typically combining all of the flags in `ServiceManagerAccess` gives you (perhaps) full access to the SC panel and with `bitflags` that we use it's easy, just call `::all()` on the struct. I could be wrong but that worked well for most manipulations that we've done in the past.

NuSkooler: `::all()` currently doesn't do the job, but perhaps if we added the missing flags. It looks like the mask set in ALL_ACCESS has some flags that don't specifically fit within the current sets though.

src/service.rs
pronebird: Private wrapper type `ScHandle` exposes `raw_handle` but a lot of places use `to_raw`. 🤷‍♂️

src/service.rs
```diff
@@ -80,6 +81,9 @@ bitflags::bitflags! {
 
         /// Can use user-defined control codes
         const USER_DEFINED_CONTROL = Services::SERVICE_USER_DEFINED_CONTROL;
+
+        /// Full access to the service object
+        const ALL_ACCESS = Services::SERVICE_ALL_ACCESS;
```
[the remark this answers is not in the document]
NuSkooler: I commented above, but it gets a little tricky here as some of those flags are outside of the "Service" flags table, such as `WRITE_DAC` which is a generic Windows security flag.

src/service.rs
NuSkooler: Would you like to switch it `to_raw`? I'm happy to do so if that's your preference!

src/service.rs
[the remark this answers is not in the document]
pronebird: I see. For what it's worth `ServiceAccess:DELETE` also points to `FileSystem::DELETE`, which is funny to say the least but it's indeed needed to be able to delete the service. I suppose the story repeats for these flags too. I guess Windows folks modeled it in such way where they thought that a common set of flags should be respected by services too. Everything is an object of some sorts I guess.

src/service.rs
pronebird: Personally I'd use `to_raw` to keep the interface homogeneous and familiar to consumers.

faern: One could argue we should stop we the bespoke methods and just implement https://doc.rust-lang.org/stable/std/os/windows/io/trait.IntoRawHandle.html for the relevant types instead. But that's not going to happen in this PR. For now I think we should just be consistent with what this library usually uses (`to_raw`)

src/service_manager.rs
faern: Thanks! This is good.

faern (approved):

faern:
@NuSkooler Thank you for the contribution!

PR #127, opened 2024-04-26
mark ffi_service_main as public

ssrlive:
so we can use `ffi_service_main` from other module.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/127)

### DIFF
--- a/src/service_dispatcher.rs
+++ b/src/service_dispatcher.rs
@@ -39,7 +39,7 @@ macro_rules! define_windows_service {
     ($function_name:ident, $service_main_handler:ident) => {
         /// Static callback used by the system to bootstrap the service.
         /// Do not call it directly.
-        extern "system" fn $function_name(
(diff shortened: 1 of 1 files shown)


pronebird:
Why would you want an implementation detail being public?

ssrlive:
Here is my code. I call `ffi_service_main` from `bin` module.

`windows_service::service_dispatcher::start("dns-over-tls", dns_over_tls::ffi_service_main)?;`

image


> Why would you want an implementation detail being public?

ssrlive:
I've changed it to pub(crate) so the function won't be accidentally exported. 
Anyway, this also adds a bit of flexibility to the user.

faern:
This does not sound like something that should be exported by default. You can very easily re-export it yourself, so you are not blocked from implementing it in the way you want to anyway.

PR #128, opened 2024-05-01
feat: write detailed IO error info in winapi call

ArcticLampyrid:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/128)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,7 +216,7 @@ impl std::fmt::Display for Error {
                 "{} contains a nul byte in element at {} index",
                 name, index
             ),
-            Self::Winapi(_) => write!(f, "IO error in winapi call"),
+            Self::Winapi(io_err) => write!(f, "IO error in winapi call: {}", io_err),
         }
     }
 }


pronebird:
Yes but why? The `Error` type exposes `source()` so for example if you use `anyhow` you can [unroll the error chain](https://docs.rs/anyhow/1.0.82/anyhow/struct.Error.html#display-representations) by using a `{:#}` format. This would also prevent duplicate messages from appearing at each level, i.e:

```
Error: IO error in winapi call

Caused by:
  No such file or directory (os error 2)
```

as opposed to:

```
Error: IO error in winapi call: No such file or directory (os error 2)

Caused by:
  No such file or directory (os error 2)
```

ArcticLampyrid:
> For example if you use `anyhow` you can [unroll the error chain](https://docs.rs/anyhow/1.0.82/anyhow/struct.Error.html#display-representations) by using a `{:#}` format. This would also prevent duplicate messages from appearing at each level.

Curious whether this is a conventional error handling method. 

I'm using [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) to wrap every error. And then there is no such magic for `thiserror` to handle `{:#}` format.

pronebird:
A bit while ago @faern [wondered the same thing](https://users.rust-lang.org/t/do-people-not-care-about-printable-error-chains-a-k-a-how-to-nicely-implement-display-for-an-error/35362/2)

I believe the correct thing to do is to leave the details of source error to the source error itself, but maybe others have a different view.

I have been carrying [this small helper](https://github.com/mullvad/mullvadvpn-app/blob/0eb08329ca692f831b603252d00bad4604a05f18/talpid-types/src/error.rs#L13-L21) to print error chains for years and I hope one day `std::error::Error` gains a method to do the same, especially given that we already have `source()` implemented as a part of `std::error::Error`

ArcticLampyrid:
Make sense.

PR #132, opened 2024-09-09
feat: handle device event in service handler

vthib:
I have recently needed to receive device events as a windows service, and found out that I couldn't with the current version of windows-service. This PR is here to make this possible.

To receive those events, a call to `RegisterDeviceNotification` must be done. This is of course not done in this crate and let instead to the user.

Then, instead of parsing the event data for every subtype of device events, I've chosen to keep the event data in raw form, and let the user of the crate do this parsing. This does not seem to be quite how the rest of the service event handler is written, but this is by far the simplest:

- The event data for device events can be quite complex (with first a HDR type, then different type of objects depending on the subtype stored in the header), and which type of events is received also depends on which parameters was provided to `RegistryDeviceNotification`
- There are several type of device events (arrival, remove, custom, etc), and the user may only be interested in a subset of those

So all in all, it's imho much simpler and also efficient to let the user handle the conversion of this event data into rust. For example in my case, I only need to handle device arrival, and only some subset of those events.

Let me know what you think, or if you would prefer parsing those events to provide a nicer experience to the user. Also, thanks for your work on this crate :)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/132)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -954,6 +954,50 @@ impl PowerEventParam {
     }
 }
 
+/// Enum describing the type of DeviceEvent event.
+///
+/// In order to receive those events, `RegisterDeviceNotification` must be called first.
+/// See <https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerdevicenotificationw>.
+///
+/// The value of each variant corresponds to the event data, which is kept in raw form as it can
+/// be quite complex. Please refer to the `lParam` documentation here for more details on this data:
+/// <https://learn.microsoft.com/en-us/windows/win32/devio/wm-devicechange>.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum DeviceEventParam {
+    DeviceArrival(*const c_void),
+    DeviceRemoveComplete(*const c_void),
+    DeviceQueryRemove(*const c_void),
+    DeviceQueryRemoveFailed(*const c_void),
+    DeviceRemovePending(*const c_void),
+    CustomEvent(*const c_void),
+}
+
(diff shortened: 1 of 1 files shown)


src/service.rs
```diff
@@ -954,6 +954,50 @@ impl PowerEventParam {
...
+///
+/// In order to receive those events, `RegisterDeviceNotification` must be called first.
+/// See <https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerdevicenotificationw>.
+///
+/// The value of each variant corresponds to the event data, which is kept in raw form as it can
+/// be quite complex. Please refer to the `lParam` documentation here for more details on this data:
+/// <https://learn.microsoft.com/en-us/windows/win32/devio/wm-devicechange>.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum DeviceEventParam {
```
pronebird, line 967: Given that each variant contains `*const c_void`, would it make sense to represent that as a struct with two fields instead?

vthib: That would be quite equivalent yes, except that it would remove the possibility to add safe wrappers for some of those events in the future

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/132)* status: 0 of 1 files reviewed, all discussions resolved

pronebird (commented):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/132)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #133, opened 2024-10-22
Bump windows-sys version to 0.59

zaddach:
This updates the windows-sys version to 0.59.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/133)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ bitflags = "2.3"
 widestring = "1"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.52.0"
+version = "0.59.0"
 features = [
     "Win32_Foundation",
     "Win32_Security",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/mullvad/windows-service-rs"
 license = "MIT OR Apache-2.0"
 edition = "2021"
 # Keep in sync with CI job in `build-and-test.yml`
-rust-version = "1.58.0"
+rust-version = "1.60.0"
 
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ widestring = "1"
 version = "0.59.0"
 features = [
     "Win32_Foundation",
-    "Win32_Security",
     "Win32_Storage_FileSystem",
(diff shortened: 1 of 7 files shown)


src/sc_handle.rs
```diff
@@ -1,15 +1,15 @@
-use windows_sys::Win32::{Security, System::Services};
+use windows_sys::Win32::System::Services;
 
 /// A handle holder that wraps a low level [`Security::SC_HANDLE`].
```
faern, line 3: Please don't forget to update the docs accordingly. Please make sure that `cargo doc` builds without warnings or errors! I have been meaning to get this check into the CI, but have not done so for this repo yet.

Cargo.toml
```diff
@@ -20,7 +20,7 @@ bitflags = "2.3"
 widestring = "1"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.52.0"
+version = "0.59.0"
 features = [
     "Win32_Foundation",
     "Win32_Security",
```
faern, line 26: Is the security feature maybe redundant now, or is it used by something else? I was wondering since we stopped using the `Security` module for `SC_HANDLE` now.

faern (changes requested):
Thanks for the upgrade! A few things to fix before merge.

Cargo.toml
zaddach: `SC_HANDLE` moved from `Win32::Security` to `Win32::System::Services`. I've tried to remove `Win32::Security` and the crate builds fine, so I've committed the change.

zaddach:
I've updated the MSRV, and tested that both `cargo test` and `cargo doc` finish without errors.

src/sc_handle.rs
faern: Ping ^

faern:
You need to update what Rust version the CI runs on. It still runs on 1.58. Please see https://github.com/mullvad/windows-service-rs/blob/main/.github/workflows/build-and-test.yml#L24-L28

Also don't forget my comment about the documentation links invalidly pointing to `Security::SC_HANDLE`

src/sc_handle.rs
zaddach: done.

zaddach:
Ahahah, I must still have been a bit tired in the morning. Forgot to actually add the files before committing. Here you go.

faern (approved):
Looks good! Thanks for the contribution!

PR #134, opened 2025-01-09
Display winapi errors

radkum:
If the WinAPI connection fails, we have no information about what went wrong. The purpose of the PR is only to add WinAPI error information to the error message

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/134)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -209,14 +209,14 @@ impl std::fmt::Display for Error {
             Self::LaunchArgumentsNotSupported => {
                 write!(f, "kernel drivers do not support launch arguments")
             }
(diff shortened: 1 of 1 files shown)


faern:
You can already obtain this information from the `source()` method on the error. The winapi error is one step down in the error chain. An error should not both return sub-errors in its `source()` method and include its error message in itself. Your PR makes the error chain redundant. If you now print the error chain it will look something like:

```
Error: IO error in winapi call: It went bad on the inside
Caused by: It went bad on the inside
```

I have been trying to get this merged as a best practice for Rust in general, but the discussion died after a while :) https://github.com/rust-lang/api-guidelines/pull/210

PR #139, opened 2025-06-12
update windows-sys to v0.61

ssrlive:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/139)


dlon:
Would you like to bump this to 0.61 and also increase MSRV (to 1.71, IIRC)?

PR #142, opened 2025-08-19
Add function to check if current process is running as a service

tyilo:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/windows-service-rs/142)


dlon:
Is running in session 0 equivalent to being a service? A subprocess launched by a service would (I assume) be considered a service according to this function.

tyilo:
Hmm, good point.

Maybe it would be ok with another name?

